PR #31, opened 2016-05-27
add key to list item

tongdu99:


hakosl:
I'm getting `ExceptionsManager.js:76Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Button`. See https://fb.me/react-warning-keys for more information.` When using a icon button, I think this commit would solve this issue. Or is it only me thats getting this error.

sankhadeeproy007:
I did try a similar solution and it broke. You can copy this piece of code over to your project and check if it works. We might take another look at it.

hakosl:
@sankhadeeproy007 @tongdu99's code obviously isn't going to work, because you have to set the key property on the JSX element, and not on the children array. The way I implemented it was just to give the icon the key="icon" and text key="text". This seems to work fine, and I don't see any reason to not do it this way when there is only two elements.

sankhadeeproy007:
Care to send a PR to fix this?

tongdu99:
sorry guys, it was an accidental pull request. I attempted to fix it but ended up writing my own control.

hakosl:
@sankhadeeproy007 I decided that I'm not going to use NativeBase anyways, but i decided to go ahead and create the pull requests anyway.

sanketsahusoft:
@hakosl May I know the reason for you deciding on not to use NativeBase?

hakosl:
@sanketsahusoft Some things weren't customisable enough(couldn't get the spacing between icon and text on the button to be just righ, i also couldn't use other font sets than ioniconst, while i wanted to use FontAwesome). I was never really interested in the layout component anyways, just interested in getting some simple components like buttons and textfields, without having to write them myself.

sanketsahusoft:
@hakosl We hear you! Thanks for the feedback. It helps us to improve it.

We are going to have a lot of customizable options and since each component of NativeBase replaces a react native component, you can always pass style property to change it's styling. We are also going to introduce composition of component for customization where in you can change any behavior of the component, in fact it will be replaceable with a component of your choice using the same interface.

We are working on how can you easily change fonts as well. And also, you can just import the components which you need.

hakosl:
@sanketsahusoft Sounds good, I will keep an eye on the project and see if i find it usefull in the future. 
Cheers

patricktan98:
The problem occurred in CardItem again.
You may replicate this issue by adding cardItem in ur project.
May I know any solution to solve tis?

nsorathia-rbi:
Facing the same issue, getting warning message: Warning: Each child in a list should have a unique "key" prop. any idea?

PR #132, opened 2016-08-23
Theme is not applied to icon.

chaitanya0bhagvan:
Theme was not applied to Icon component. iconFontFamily always defaulted to Ionicons


himanshu-sahusoft:
Thanks @chaitanya0bhagvan

Issue #137, opened 2016-08-24
icon.props is undefined. 


zibs:
Hi all,

I've seen this error pop up in a couple of different places for various reason. I'm thinking it has something to do with trying to render dynamic content inside of a button/input group specifically. 

``` javascript
<Button block bordered> {"this doesnt work"} </Button>
```

This will throw an icon.props undefined error. 

The same thing happens when I try to render an icon depending on if it should be rendered or not. The detectIconToRender function would return null if there is no function to render

``` javascript
<InputGroup borderType='underline' >
      {this.detectIconToRender()} 
   <Input placeholder='Dynamic Icons' />
</InputGroup>
```

As mentioned, both of these cases return an icon.props undefined error. 

I've looked in the code and it seems that dynamically rendered content is getting interpreted as iconLeft as it attempts to render the content of the button/input group. 

Thanks!

sankhadeeproy007:
We too have come across this bug and it's annoying. Will fix it sometime in the near future.

sankhadeeproy007:
Fixed with v0.5.20

JulianKingman:
I'm getting this issue with a button that has iconLeft, I'm on 0.5.22, here's what it looks like (using react-native-vector-icons/MaterialIcons):
````javascript
<Button block transparent large iconLeft onPress={this.toggleDatePicker}>
  <Icon name="event"/>
  {this.formatDate(this.state.selectedDate)}
</Button>
````
Stripping it down, I still get the error with just this:
````javascript
<Button iconLeft>
  <Icon name="event"/>
  <Text>Button</Text>
</Button>
````

PR #144, opened 2016-08-26
Removing null or not valid child components from props.children

redbaron76:
This fix removes not valid or null children when creating a new childrenArray from props.children


sankhadeeproy007:
Hi, thanks for helping out. Although, I didn't have the time to go through the code and test it out, I assume it's a fix for https://github.com/GeekyAnts/NativeBase/issues/137

Issue #142, opened 2016-08-26
Using custom shortcut components


redbaron76:
Hello there. I've built a `NavBar` "shortcut" component that basically builds the `Header` because imho all this structure is great when you want to describe a scene but becomes quickly verbose when you have to repeat it over and over in all the scenes of the project. And in my case, using Redux as state container, would be faster to manage the `Header` in a more programmatic way.

The component's signature is as follows:

```
NavBar.propTypes = {
    title:React.PropTypes.string,
    left:React.PropTypes.shape({
        icon: React.PropTypes.string,
        label: React.PropTypes.string,
        onPress:React.PropTypes.func,
    }),
    right:React.PropTypes.shape({
        icon: React.PropTypes.string,
        label: React.PropTypes.string,
        onPress:React.PropTypes.func,
    })
};
```

and it renders the `Header` component with three custom methods for left/right buttons and the title

```
render() {
        return (
            <Header>
                {this._renderLeftButton()}
                {this._renderTitle()}
                {this._renderRightButton()}
            </Header>
        );
    }
```

but, of course, looking at the code, the `Container` component expects an explicit `Header` component as one of its children and it doesn't render it. (Isn't possibile to avoid this behavior? If it's the Header, render it as expected otherwise just render the "unexpected" component... - up to developer what is going on there)

I noticed also that the `Container` component is just a `View` with `flex:1`... so I replaced the `Container` with a simple `<View style={{flex:1}}>` and now my custom `NavBar` component works **BUT** I have to return always something from `this._renderLeftButton()` and `this._renderRightButton()` (can't return null i.e.) otherwise I got an `Cannot read property 'props' of undefined`... so I basically can't decide when and where showing my buttons.

Any suggestion about this? Am I missing something?

This is a great project, well documented and a big help for developers but probably, in my opinion, it lacks a little bit in terms of flexibility, especially in projects one step above.

dunnock:
Same, I would like to customize Icons in the code like below:

```
const errorStyle = {color: "#bb0033"}
const validStyle = {color: "#33bb00"}
const ValidationIcon = (props) => <Icon
                  key={props.key}
                  name={ props.text === "" ? 'ios-alert' : 'ios-checkmark-circle'}
                  style={ props.text === "" ? errorStyle : validStyle }
                />

```

flyandi:
+1 on this. It doesn't seem to work without rewriting the Container element.

sanketsahusoft:
[...]

PR #145, opened 2016-08-26
Allow role based custom Header and Footer components

redbaron76:
This allows to replace Header and Footer components into Container with a custom 'shortcut' component with a specific and mandatory role 'header' or 'footer' as described in GitHub issue #142

Example:

```

const left = {
   icon: "ios-arrow-back",
   label: "Back",
   onPress: () => alert('Back')
};

const right = {
   icon: "ios-menu",
   onPress: () => alert('Menu')
};

return (
   <Container>

      <CustomNavBar role="header" title="Home Page" left={left} right={right} />

      <Content>
         // Scene content
      </Content>

   </Container>
);
```

The `CustomNavBar` will accept props and will render like this:

```
...

   render() {
        return (
            <Header>
                {this._renderLeftButton()}
                {this._renderTitle()}
                {this._renderRightButton()}
            </Header>
        );
   }

CustomNavbar.propTypes = {
    title:React.PropTypes.string,
    left:React.PropTypes.shape({
        icon: React.PropTypes.string,
        iconPos: React.PropTypes.string,
        label: React.PropTypes.string,
        onPress:React.PropTypes.func,
    }),
    right:React.PropTypes.shape({
        icon: React.PropTypes.string,
        iconPos: React.PropTypes.string,
        label: React.PropTypes.string,
        onPress:React.PropTypes.func,
    })
};
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,11 @@
 .AppleDouble
 .LSOverride
 
+.idea
+
 # Icon must end with two \r
-Icon
+Icon
+
 
 # Thumbnails
 ._*
--- a/Components/Widgets/Button.js
+++ b/Components/Widgets/Button.js
@@ -13,183 +13,187 @@ import _ from 'lodash';
 
 export default class Button extends NativeBaseComponent {
 
-   propTypes: {
-      style : React.PropTypes.object,
-      textStyle : React.PropTypes.object,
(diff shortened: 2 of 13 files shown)


shivrajkumar:
NativeBase is restructured with v2.x and this is fixed

PR #218, opened 2016-10-04
Allow extra content to be loaded within container

th3no0b:
Allow extra content to be loaded within <Container /> component if an 'allowExtraContent' prop is passed as true.

Useful for things like Bottom Sheets and popups that are positioned absolutely

### DIFF
--- a/Components/Widgets/Container.js
+++ b/Components/Widgets/Container.js
@@ -16,6 +16,23 @@ export default class Container extends NativeBaseComponent {
 	propTypes: {
         style : React.PropTypes.object
     }
+	
+	renderOther(){
+		if(Array.isArray(this.props.children)) {
+
+			return _.filter(this.props.children, function(item) {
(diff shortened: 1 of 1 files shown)


pcfutures:
+1 This would be helpful

Components/Widgets/Container.js
```diff
@@ -16,6 +16,23 @@ export default class Container extends NativeBaseComponent {
 	propTypes: {
         style : React.PropTypes.object
     }
+	
+	renderOther(){
+		if(Array.isArray(this.props.children)) {
+
+			return _.filter(this.props.children, function(item) {
+				if(item && (item.type != Header && item.type != ViewNB && item.type != Content && item.type != Image && item.type != View && item.type != ScrollView && item.type != Footer)) {
```
EricMFischer: You can use _contains from lodash, e.g. `_.contains([1, 2, 3], 3);` returns true

shivrajkumar:
NativeBase is restructured in v2.x.
Any react native component can be included within the container now

PR #219, opened 2016-10-04
create new object instead of setting the padding directly on props in Content component

ansgarm:
Hey,

I ran into crashes as my props were somehow converted into immutable objects. Maybe there are some performance things happening in newer versions of react-native and thus this error hasn't surfaced before.

Anyways, here's a pull request that doesn't directly modify the props object and allows overwriting of the padding by the user :)

### DIFF
--- a/Components/Widgets/Content.js
+++ b/Components/Widgets/Content.js
@@ -34,8 +34,8 @@ export default class Content extends NativeBaseComponent {
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


david50407:
But this will copy all content of `this.props.contentContainerStyle` that misses `StyleSheet`'s caching, better way is using arrayed-styles, see my issue https://github.com/GeekyAnts/NativeBase/issues/227

ansgarm:
Ah, sounds good!

PR #226, opened 2016-10-07
ListItem : Thumbnail + iconRight support

MatthieuLemoine:
Allows to use both Thumbnail and iconRight in a ListItem.

``` html
<ListItem iconRight>
  <Thumbnail size={50} source={{uri : avatarUrl}}/>
  <Text>Username</Text>
  <Text note>Location</Text>
  <Icon name="ios-arrow-forward"/>
</ListItem>
```


shivrajkumar:
Works in NativeBase v2.x

PR #247, opened 2016-10-20
fixed android button issue

### LINKED ISSUE #201: Button issue with mutable text from state on Android - undefined is not a function
nb v0.5.8

`<Button primary onPress={this.toggleState} style={styles.buttons}> {this.state.submitButtonText} </Button>`

error in button.js (line 188)

`return <Text style={this.getTextStyle()}>{(Platform.OS==='ios') ? this.props.children : this.props.children.toUpperCase()}</Text>`

if I remove the .toUpperCase() for Android it runs fine, but with this code i get an "undefined is not a function" error. 

Thanks for your help
java99

ios version is running without issue in both cases

jaronoff97:
I fixed #201 by following the suggestion in the comments, i'll be shipping an app soon and it would be best if i didn't have to edit anything in my node modules.

### DIFF
--- a/Components/Widgets/Button.js
+++ b/Components/Widgets/Button.js
@@ -191,7 +191,7 @@ export default class Button extends NativeBaseComponent {
                 }
             }
             else {
-                return <Text style={this.getTextStyle()}>{(Platform.OS==='ios') ? this.props.children : this.props.children.toUpperCase()}</Text>
+                return <Text style={this.getTextStyle()}>{(Platform.OS==='ios' || !this.props.capitalize) ? this.props.children : this.props.children.toUpperCase()}</Text>
             }
 
             return newChildren;


sankhadeeproy007:
@himanshu-satija Please review this.

himanshu-satija:
This fails if I pass capitalize prop with the Button.

In the original issue, it failed due to this.props.children being an array 
`<Button onPress={this.props.turnOff}> {`close ${this.props.id}`} </Button>`

This causes the children prop to be an array[`' ', 'close 1', ' '`] because of the extra spaces outside curly braces. 
This should work fine.
`<Button onPress={this.props.turnOff}>{`close ${this.props.id}`}</Button>`

Actually, we should handle this case in a different way. Feel free to send another PR for that. Or we can fix that in the next version.

jaronoff97:
@himanshu-satija where is that code you put? i grepped my fork and couldn't find it.

himanshu-satija:
This is not in your code. It is the Button usage which is causing this issue
https://github.com/GeekyAnts/NativeBase/issues/201#issuecomment-250681199

jaronoff97:
So what do you propose we do? We could do a filter for anything that isn't a component at the beginning for the method.

himanshu-satija:
Yes, that would be the best approach

jaronoff97:
Awesome, I'll make the necessary changes and submit another PR

himanshu-satija:
Thanks :)

jaronoff97:
Is there anything else I should filter for other than a space, such as null, undefined, etc?

himanshu-satija:
Sure, we can filter undefined and null elements

jaronoff97:
Are there any automated tests for the lib?

himanshu-satija:
No, but it is there on our roadmap.

PR #250, opened 2016-10-22
Fix Header when a `null` child is being passed

mweibel:
I had a bug when using the Header in the following way:

```
const someVar = false;
let someButton = null;
if (someVar === true) {
  someButton = <Button><Icon name='ios-arrow-back' /></Button>;
}

return (
  <Header>
    {someButton}
    <Title>Test</Title>
  </Header>
);
```

Issue is that the Component Header checks if the children is an array or not. If not it will just immediately return the children to be rendered. Otherwise it will try to layout the children accordingly to what is being passed. 

Before I started fixing the issue I however was not very comfortable with changing anything due to the code quality and no tests being around. 
Therefore I first added tests for the use cases I found in the documentation, using [Jest](https://facebook.github.io/jest/). This has been done in https://github.com/GeekyAnts/NativeBase/commit/c6491a7252f922b09611b119f25b6d65ed40a51b.

After adding the tests I started refactoring the code. Indentations were wrong (if else clauses where indentation was completely off), else clauses where no else clause was needed because you can return early instead. 
Also everything in one method is too hard to follow as well. 
I therefore refactored the existing component into what's visible in https://github.com/GeekyAnts/NativeBase/commit/4fefbb00147ce1291d3319180dac53034f3af791. 

After doing that, fixing the actual bug has been easy: https://github.com/GeekyAnts/NativeBase/commit/b62a0ae986c7ee203cb3253381e6c64201a6f6a8. 

The refactored code is much easier to read and to make changes. It is however not yet optimal. The styles which I added on top (instead of being in each rendered component directly) is reducing duplication but I think there's still a lot of optimizations one could do.

I do strongly suggest to add tests for all other components as well and then start to refactor it in a similar way. This will make future changes much easier.
Additionally I recommend using [Travis-CI](http://travis-ci.org/) for running the tests automatically for every change and pull request. 

Also I've removed lodash in this specific component. If you only use `_.remove` and other similar functions from lodash, I'd recommend using standard javascript functions (e.g. the ones in Array, like `Array.prototype.filter`) instead.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["react-native"]
+}
--- a/Components/Widgets/Header.js
+++ b/Components/Widgets/Header.js
@@ -1,8 +1,8 @@
 /* @flow */
 'use strict';
 
-import React from 'react';
-import { Platform} from 'react-native';
+import React, { cloneElement } from 'react';
+import { Platform } from 'react-native';
 import NativeBaseComponent from '../Base/NativeBaseComponent';
 import computeProps from '../../Utils/computeProps';
 import Button from './Button';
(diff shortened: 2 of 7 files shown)


mandre42:
+1 :)

mweibel:
Will this get merged at some point? I'd resolve the merge conflict then, otherwise I won't do that.

Issue #256, opened 2016-10-27
FooterTab returns error when using an Icon


jesouhaite08:
I used the [docs](http://nativebase.io/docs/v0.5.9/components#footerTabs) as a base to create a FooterTab. 

I receive the following error if I add an Icon element to a button: undefined is not an object (evaluation 'child.props.children.props.name') from FooterTab.js:114 (has since changed, now [line 142](https://github.com/GeekyAnts/NativeBase/blob/master/Components/Widgets/FooterTab.js#L142)).

Below is the code I am using in my render function:

```
 <Footer>
          <FooterTab>
              <Button active={this.state.detailsTab} onPress={() => this.toggleDetailsTab()}>
                  Details
                  <Icon name='ios-home' style={globalStyles.inspectionsItemIcon} />
              </Button>
              <Button active={this.state.inspectionTab} onPress={() => this.toggleInspectionTab()}>
                  Inspection
              </Button>
              <Button active={this.state.insectsTab} onPress={() => this.toggleInsectsTab()}>
                  Insects
              </Button>
              <Button active={this.state.radonTab} onPress={() => this.toggleRadonTab()}>
                  Radon
              </Button>
          </FooterTab>
 </Footer>
```

If I remove the Icon from the first button everything works well.

Any ideas? Thanks in advance.

sankhadeeproy007:
I'm curious to know, why would you want Icon in only a single tab? You'd rather have them in all the tabs or neither right?

jesouhaite08:
@sankhadeeproy007 I would like to have an icon in all tabs. I just put the icon in one for the sake of this example. I had removed all of them so that I could continue working.

monodop:
I am having this issue as well, but only if my container is wrapped in a react-native-drawer

monodop:
Looks like this is caused by the lodash _.remove modifying your Button's children props. The above PR should fix it, or you can take a look at the PR to apply it right now as a quick fix. It worked for me, so hopefully it's the correct fix to this bug.

sankhadeeproy007:
@jesouhaite08 silly question, but did you try putting an Icon inside all the buttons? I remember having tested it throughly. #242 even has a working example of the same.

jesouhaite08:
@sanketsahusoft It produces the same error with an Icon inside all of the buttons.

jesouhaite08:
@monodop Mine is only wrapped inside of a Container. Weird. Thanks for the reply.

[...]

PR #263, opened 2016-10-31
Fix modification of props #256

monodop:
The FooterTab component modifies the children's props in place using _.remove, which causes the children array to be broken on subsequent renderings. This fix simply shallow clones the children props before making changes to it, fixing this issue #256 . Sorry if this isn't exactly right, but it should be the right direction at least. Let me know if there's any issues with it :)

### DIFF
--- a/Components/Widgets/FooterTab.js
+++ b/Components/Widgets/FooterTab.js
@@ -73,7 +73,8 @@ export default class Footer extends NativeBaseComponent {
         var newChildren = [];
 
         {childrenArray.map((child, i) => {
-            if (typeof child.props.children==='string') {
+        	let children = _.clone(child.props.children);
+            if (typeof children==='string') {
                 newChildren.push(React.cloneElement(child, {
                     style: [this.getInitialStyle().btnStyle, {backgroundColor: (child.props.active) ? this.getTheme().tabActiveBgColor : undefined }],
                     vertical: true,
@@ -84,13 +85,13 @@ export default class Footer extends NativeBaseComponent {
             }
             else {
                 let iconElement = [];
-                iconElement = _.remove(child.props.children, function(item) {
+                iconElement = _.remove(children, function(item) {
                     if(item.type == IconNB) {
                         return true;
                     }
                 });
                 let badgeElement = [];
-                badgeElement = _.remove(child.props.children, function(item) {
+                badgeElement = _.remove(children, function(item) {
                     if(item.type == Badge) {
                         return true;
                     }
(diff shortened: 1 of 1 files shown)


sankhadeeproy007:
Not sure sure why removing the Icon element should be a problem because we are rendering the Icons separately [here](https://github.com/monodop/NativeBase/blob/master/Components/Widgets/FooterTab.js#L108) anyways.

monodop:
Right, but what happens is that subsequent renders to the element can have modified props from what you originally passed. You should really never directly modify props or state.

monodop:
Some example scenarios with some comments:

This works as expected:
```es6
import React, { Component, PropTypes } from 'react';
import { Text, StyleSheet } from 'react-native';
import { Container, Button, Content, Footer, FooterTab, Icon } from 'native-base';

export default class MyScene extends Component {
    constructor(props) {
        super(props);
    }
    render() {
        return (
            <Container>
                <Content>
                    <Text>Sample Text</Text>
                </Content>

                <Footer>
                    <FooterTab>
                        <Button>
                            Test 1
                            <Icon name="md-menu" />
                        </Button>
                        <Button>
                            Test 2
                            <Icon name="md-more" />
                        </Button>
                    </FooterTab>
                </Footer>
            </Container>
        )
    }
}
```

However, adding a Drawer around it, does not..
```es6
import React, { Component, PropTypes } from 'react';
import { Text, StyleSheet } from 'react-native';
import { Container, Button, Content, Footer, FooterTab, Icon } from 'native-base';
import Drawer from 'react-native-drawer';

export default class SignInScene extends Component {
    constructor(props) {
        super(props);
    }
    render() {
        return (
            <Drawer content={<Text>Test Drawer</Text>}>
                <Container>
                    <Content>
                        <Text>Sample Text</Text>
                    </Content>

                    <Footer>
                        <FooterTab>
                            <Button>
                                Test 1
                                <Icon name="md-menu" />
                            </Button>
                            <Button>
                                Test 2
                                <Icon name="md-more" />
                            </Button>
                        </FooterTab>
                    </Footer>
                </Container>
            </Drawer>
        )
    }
}
```

If we step through, we'll find that an inital render is performed, and there is no error on the first render. Each button has 2 children: the text (Test 1 or Test 2), and the Icon. However, the second render, since the props have been modified, the Button only has one child, the text, but it is still an array of 1, which breaks since your statements don't account for that (they only check for a string, but that's not the real problem here). As a result, the render fails, but since the previous render succeeded it kind of doesn't matter (though you get nasty errors).

After applying the proposed fix, the props do not change between renders and instead we get a successful render each time through.

Hopefully this demonstrates my point and allows you to reproduce the problem it is attempting to solve.

himanshu-satija:
@monodop Thanks for the detailed explanation

PR #271, opened 2016-11-05
Fix BadgeNB propTypes definition.

admmasters:
A number of the propTypes definitions are malformed.


himanshu-satija:
Will fix propTypes definition in all the components in next version

PR #287, opened 2016-11-12
Add typings

wcandillon:
From https://github.com/mrphu3074/typed-native-base (cc @mrphu3074)


sanketsahusoft:
@wcandillon This is amazing! Let us go through this.

PR #291, opened 2016-11-14
Fix badge for high resolutions

priezz:
Take provided line height into consideration


himanshu-satija:
For demo code on NativeBase docs, it causes the badges to look like this 

simulator screen shot nov 18 2016 5 08 08 pm

Please modify the PR, and an explanation for what it fixes for which use case would help :)

Issue #295, opened 2016-11-15
Piker in list/forms


alejonext:
Look, the `Piker`, is a `input`, bat is determine a `View`. Well... That's is the code..
```html
<ListItem>
	<InputGroup>
		<Picker
			inlineLabel
			iosHeader="Select number"
			label="Select number"
			mode="dropdown"
			selectedValue={this.state.num}
			onValueChange={num => this.setState({ num })}>
			<Item label='One' value={1} />
			<Item label='Two' value={2} />
			<Item label='Thre' value={3} />
		</Picker>
	</InputGroup>
</ListItem>
```
And not show the piker. Is resolve.. This in [line](https://github.com/GeekyAnts/NativeBase/blob/master/Components/Widgets/ListItem.js#L181)

```javascript
if(child && ( child.type == InputGroup || child.type == Picker ) )
```

SupriyaKalghatgi:
@alejonext Take a look at [Form](https://github.com/GeekyAnts/NativeBase-KitchenSink/blob/master/js/components/form/index.js) of [NativeBase-KitchenSink](https://github.com/GeekyAnts/NativeBase-KitchenSink)

Similar to your use case

Papipo:
I still don't know how to properly render a Picker with a Label. What's InputGroup? Why isn't it documented? :(

Issue #89, opened 2016-08-01
stackedLabel/inlineLabel support for Picker


Trenrod:
Currently Picker can be used as ListItem but cannot add a label.

sankhadeeproy007:
@Trenrod can tell me what exactly your use case is and how you want it to look?

ericvan76:
I'm not sure whether my one is the same issue, try to make a form with following fields: 

- Name (A TextInput with a Icon and Label)
- Gender (A ListItem with a Icon and Label, and click it will show Picker)
- Other Switches or CheckBoxes with Icon and Label

It is very difficult to make them in a same list with same styles. 

To simplify the question, how can I use a picker with a form? the picked value should be displayed as the same style of the TextInput.

SupriyaKalghatgi:
@ericvan76 You can check the demonstration of [Form](https://github.com/GeekyAnts/NativeBase-KitchenSink/blob/master/js/components/form/index.js) in [NativeBase-KitchenSink](https://github.com/GeekyAnts/NativeBase-KitchenSink)

PR #308, opened 2016-11-20
Support use of Picker inside an InputGroup in a form

chaitanya0bhagvan:
Present implementation does not work well with a Picker inside a form. The styling does not match with the styling of other form elements containing Input. This PR also resolves issues [Issue 1](https://github.com/GeekyAnts/NativeBase/issues/295) and [Issue 2](https://github.com/GeekyAnts/NativeBase/issues/89)


chaitanya0bhagvan:
@himanshu-satija Any update on this?

PR #310, opened 2016-11-21
Gravatar component added

ozanturksever:
Code mostly modified from Thumbnail component.


himanshu-satija:
@ozanturksever This is awesome. Thanks!

PR #312, opened 2016-11-22
Fixes `DeckSwiper` changes from `57a2b0f`

jmfirth:
Something went awry in 57a2b0f.  This attempts to restore intention.  (Added code replaces old; prettification)


jmfirth:
Fixed in 4509a2b

Issue #319, opened 2016-11-24
Cannot find module 'react/lib/ReactNativePropRegistry' from ...


msom:
...
- computeProps.js
- Button.js

with react `15.4.1`

`var ReactNativePropRegistry = require('ReactNativePropRegistry');` seems to work

tanmynguyen:
hjc. Me too. so I use react 15.3.2. Admin @nativebase please fix it. Thanks

tanmynguyen:
I using native-base: 0.5.15

jasonbahl:
+1

jaisonveneri:
+1

Deepankar01:
+1

matthieugayon:
+1

helloncanella:
+1

ujwal-setlur:
+1

ottoo:
+1

ujwal-setlur:
Replacing with 

import ReactNativePropRegistry from 'react-native/Libraries/Renderer/src/renderers/native/ReactNativePropRegistry';

worked, but ReactNativePropRegistry is an internal API, I believe, and not public. Can this call be replaced with a public API?

Can this be resolved quickly please?

ujwal-setlur:
Actually, the following work:

import { ReactNativePropRegistry } from 'react-native';

OR

var ReactNativePropRegistry = require{'react-native'};

So, it seems that ReactNative is indeed publicly exposed now, but it should not be referenced with an absolute path.

This needs to be fixed in:

- NativeBase
Button.js
computeProps.js

- react-native-easy-grid
computeProps.js

ujwal-setlur:
I have created a pull request for this.

ujwal-setlur:
Sorry, my comment earlier was wrong. The explicit path still has to be specified. I saw that being checked in.

streamlesswaveson:
+1

dharness:
I encountered this error, but it seems to have stuck around even after removing native-base?

I even tried using `react-native init <new project>` to create an entirely new project -- but the error persists? Any insight on why this might be?

The path the error is referencing doens't even exist..

![](https://puu.sh/sx2qq/8c7228c6f8.png)

sankhadeeproy007:
Fixed with v0.5.16

tzyhhaur:
Seems that error still persist. However, the issue is with the package react-native-easy-grid. It has to be updated as well.

ujwal-setlur:
[...]

Issue #318, opened 2016-11-24
RN 0.38.0 ReactNativePropRegistry path changed


alexicum:
"react-native": "^0.38.0", 

`import ReactNativePropRegistry from 'react/lib/ReactNativePropRegistry';`
should be changed to
`import ReactNativePropRegistry from 'react-native/Libraries/Renderer/src/renderers/native/ReactNativePropRegistry';
`
in:
> native-base\Components\Widgets\Button.js
> native-base\Utils\computeProps.js
> react-native-easy-grid\Utils\computeProps.js

priezz:
Better to change to
```
import ReactNativePropRegistry from 'ReactNativePropRegistry';
```

razagill:
@alexicum I think the issue is still there, can we reopen it?

alexicum:
Yes!
And that is why:

> Requiring module 'ReactNativePropRegistry' by name is only supported for debugging purposes and will BREAK IN PRODUCTION!

alexicum:
#319

ujwal-setlur:
I believe this is a private API and should not be relied on. Can native-base switch to a public API instead?

ujwal-setlur:
I have created a pull request for this.

sankhadeeproy007:
Fixed with v0.5.16

SupriyaKalghatgi:
Fixed with latest release [0.5.18](https://github.com/GeekyAnts/NativeBase/releases/tag/v0.5.18)

thundenilla:
@priezz curious, where should i change this
 'import ReactNativePropRegistry from 'ReactNativePropRegistry';
since im not even using it anywhere. thanks

priezz:
@thundenilla, it is about the NativeBase sources. 'Components/Widgets/Button.js' and 'Utils/computeProps.js'. However, the paths are already fixed in the recent version, so it is just enough to update the package.

Anyway... Can someone explain why to use private React/RN API?

SupriyaKalghatgi:
@thundenilla You need not `'import ReactNativePropRegistry from 'ReactNativePropRegistry';`, since the issue has been fixed.
Have proper version of NativeBase with RN

alejonext:
Is not resolve in the version release 0.5.18.

``` 
"dependencies": {
    "color": "^0.11.4",
    "moment": "^2.16.0",
    "native-base": "0.5.18",
    "node-qs-serialization": "0.0.2",
    "numeral": "^1.5.3",
[...]

PR #324, opened 2016-11-27
Changed import of ReactNativePropRegistry to be not dependent on path.

ujwal-setlur:
This is to address issues #318 and #319. Import is now not using absolute path.


ujwal-setlur:
This doesn't quite work yet, rescinding pull request.

PR #364, opened 2016-12-18
Add missing function in Drawer

Vivihung:
If you have custom Navigator in react native TypeScript app, sometimes we need to call
`this.drawer.open();`
`this.drawer.close();`

Therefore, I add them into index.d.ts file

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -718,7 +718,10 @@ declare module 'native-base' {
     /**
      * NativeBase.Drawer
      */
-    export class Drawer extends React.Component<NativeBase.Drawer, any> { }
+    export class Drawer extends React.Component<NativeBase.Drawer, any> { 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -718,7 +718,10 @@ declare module 'native-base' {
     /**
      * NativeBase.Drawer
      */
-    export class Drawer extends React.Component<NativeBase.Drawer, any> { }
+    export class Drawer extends React.Component<NativeBase.Drawer, any> { 
+        close: Function;
+        open: Function;
+    }
```
ericvan76: Does the native-base Drawer have open()/close() methods? I know the `_root` does.
btw, there is a prop called "open: boolean" from underneath 'react-native-drawer', which actually controls the open/close of the drawer.

PR #380, opened 2016-12-24
Add missing properties in index.d.ts

### LINKED ISSUE #373: Picker.Item doesn't exist in the TypeScript definition
The TypeScript compiler can't find Picker.Item, resulting in compile time error, even though the value apparently exists.

wcandillon:
fixes #373

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,3 @@
-/**
- * Implementation of NativeBase v0.5.13
- */
 declare module 'native-base' {
     namespace NativeBase {
         /**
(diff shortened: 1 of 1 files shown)


sankhadeeproy007:
Hi @wcandillon, I think the Subtitle component is missing from the definition. Could you please take a look at that?

wcandillon:
@sankhadeeproy007 here it is: https://github.com/GeekyAnts/NativeBase/pull/428

wcandillon:
@sankhadeeproy007 Did you look at #397?

PR #397, opened 2017-01-02
Add support for custom fonts

wcandillon:
Add `family` property to the `Icon` component in order to use custom fonts via the `createIconSet` function of https://github.com/oblador/react-native-vector-icons


sankhadeeproy007:
@wcandillon We missed this! Will merge it sometime next week. Thanks!

shivrajkumar:
@wcandillon This will definitely help in using more that one Icon fontFamily.
But this PR is incomplete. To implement the above, both https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/Icon/index.js and https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/IconNB.js needs to be modified.
We will look into it and try to implement this.
Or you can send a PR implementing the same.

andidev:
Was this closed due to inactivity? Or was this feature added in latest master?

Issue #400, opened 2017-01-03
Enhancement: Better compatibility with Jest test framework


dikarel:
This library is [incompatible](https://github.com/GeekyAnts/NativeBase/issues/396) with React Native's [default test framework](https://facebook.github.io/jest) out of the box. The workaround requires the developer to tweak an obscure [configuration value](https://facebook.github.io/jest/docs/configuration.html#transformignorepatterns-array-string).

I'd like to propose either of the following enhancements to make first-time usage painless:

1. Transpile ES6 `import` statements into CommonJS-style `require` statements before publishing the package
2. Add an easily-accessible FAQ/guide for setting up tests with the Jest framework

The first is a common operation already done by other NPM packages; the second should be trivial to implement.

PR #403 is an attempt at (1)

sankhadeeproy007:
Thanks! Added to 0.5.21

Issue #396, opened 2017-01-02
Jest support


wsmlby:
Mentioned in:
https://github.com/GeekyAnts/NativeBase/issues/272

jest is currently broken with native-base
## react-native, react and native-base version
0.38.1, 15.4.1, 0.5.18
## Expected behaviour
npm test works
## Actual behaviour
npm test failed with 

```
import Drawer from './Components/vendor/react-native-drawer';
    ^^^^^^
    SyntaxError: Unexpected token import
      
      at transformAndBuildScript (node_modules/jest-runtime/build/transform.js:320:12)
```

dikarel:
**Try the following solution:**

Add the following property to your `package.json`'s `jest` config section:

```
"transformIgnorePatterns": [
  "/node_modules/(?!native-base)/"
]
```

Like so:

![screen shot 2017-01-03 at 8 31 53 pm](https://cloud.githubusercontent.com/assets/15131271/21609206/b2a6ceca-d1f3-11e6-9917-7fccaf5efcc4.png)

**Explanation**

1. By default, Jest [does not attempt to transpile](https://facebook.github.io/jest/docs/configuration.html#transformignorepatterns-array-string)  3rd party libs (i.e. anything under `node_modules`). This makes sense, as most "proper" npm packages would do the transpilation already during install or before publishing as a package.

2. The config above changes Jest behaviour from "don't transpile any 3rd party libs" to "don't transpile any 3rd party libs, _except for native-base_"

3. Your test should take more time to run as Jest will now have to transpile native-base.

4. The proper solution is to have native-base transpile ES6 code upon install or before publishing, so that we don't have to deal with this workaround

sankhadeeproy007:
This has been added to 0.5.21. Please update and let us know if that works.

jdnichollsc:
Hi guys,
One question... Is possible to detect if the drawer is opened?
I'm getting issues with the following tests (Using enzyme):

1. Using **spy** to detect if the function was executed
```
describe('when the filters Button are pressed', () => {
  const wrapper = mount(<ProductList {...testProps} />)

  it('should display a side menu to select filters', () => {
    wrapper.find('[className="filtersBtn"]').first().props().onPress()
    
    const spy = jest.spyOn(wrapper.find(Drawer).first().props(), 'onOpen')

    expect(spy).toHaveBeenCalled()
  })
})
```
**Error:** `TypeError: Cannot read property '_isMockFunction' of undefined`

[...]

PR #403, opened 2017-01-03
Add transpilation step to ensure native-base works with Jest out of the box

dikarel:
**Changes**

- Add `npm run compile` command to transpile source into `dist` directory
- `npm publish` publishes the `dist` directory, the `Fonts` directory, and `index.d.ts`
- Ignore extraneous files on publish (images, config files, dotfiles)

Fixes the following issues: #396, #400

**Testing**

- Published as [native-base-transpiled](https://www.npmjs.com/package/native-base-transpiled) and used locally on a React Native project
- Verify via [unpkg](https://unpkg.com/native-base-transpiled@0.5.25/) that correct files are published

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,7 @@
 {
   "sourceMaps": true,
   "presets": [
+    "es2015",
     "react"
   ],
   "plugins": [
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,10 @@
+{
+  "sourceMaps": true,
+  "presets": [
+    "react"
+  ],
+  "plugins": [
+    "transform-class-properties",
+    "transform-object-rest-spread"
+  ]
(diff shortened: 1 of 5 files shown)


sankhadeeproy007:
One small problem with this PR is that the `Fonts` directory doesn't get published with the package. Any idea why that might be happening?

dikarel:
Thanks, good catch.

I forgot to add the fonts directory into package.json's "files" property.

Will send out a fix within 24h

dikarel:
@sankhadeeproy007: I've made sure the fonts dir also gets published. You can verify below:

https://unpkg.com/native-base-transpiled@0.5.25/

FDiskas:
Probably same goes to `native-base-shoutem-theme` dependency

Issue #511, opened 2017-02-14
ejectTheme message incorrect, import path no longer valid


mribbons:
## react-native, react and native-base version
react@15.4.2
react-native@0.41.2
native-base@2.0.6

## Expected behaviour
When we run
```node node_modules\native-base\ejectTheme.js```

The message is incorrect.

https://github.com/GeekyAnts/NativeBase/blob/master/ejectTheme.js#L13

## Actual behaviour
Either there should be an index.js in src\theme, or the L13 above should be:
```    'import getTheme from \'./native-base-theme/components';',```

## Is the bug present in both ios and android or in any one of them?
Tested android, must also affect ios

sankhadeeproy007:
The idea is to eject both the components as well as the variable folders. It seems to work fine. What exactly is the problem at your end? Are the folders not getting copied?

mribbons:
Hi,
The folders are copied, the issue is that the eject script message is incorrect:

```import getTheme from './native-base-theme';```

There is no longer an index.js in ./native-base-theme, so the import fails. 

In order to make it work, one has to import from the components folder:

```import getTheme from './native-base-theme/components';```

sankhadeeproy007:
Oops! Sorry. Will be fixed.

srhise:
There is still an issue with the docs, it references the wrong path in the 2.0.0 Customize Docs. It threw me off: 

![image](https://cloud.githubusercontent.com/assets/914113/23487534/c095e41a-feb4-11e6-8c19-916abb6555af.png)

mrezzaaa:
@srhise i think there is no `import getTheme from ` inside `native-base-theme/components` 
@sankhadeeproy007

PR #514, opened 2017-02-14
Resolves https://github.com/GeekyAnts/NativeBase/issues/511

mribbons:
Hi @sankhadeeproy007 

https://github.com/GeekyAnts/NativeBase/commit/93565b5ca3b9d2049a8cd8223561de6abd7272bc didn't quite do the job, because it advises the consumer to import ./native-base-theme/index which doesn't exist and resulted in an error:

ResolveError: Unable to resolve module ./native-base-theme/index

### DIFF
--- a/ejectTheme.js
+++ b/ejectTheme.js
@@ -10,7 +10,7 @@ try {
     'NativeBase theme has been copied at ' + path.join(process.cwd(), 'native-base-theme'),
     'Here\'s how to theme your app',
     '',
-    'import getTheme from \'./native-base-theme/index\';',
+    'import getTheme from \'./native-base-theme/components\';',
     'export default class ThemeExample extends Component {',
     'render() {',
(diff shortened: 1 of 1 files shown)


sankhadeeproy007:
Ah yes, should've been `components`. Anyways we would have tested it before publishing and fixed it. But yeah thanks!

mribbons:
Thanks.

Actually the reason I came across it was I was debugging another issue and decided to npm un/install native-base again.

I went from 2.0.4 to 2.0.6 and encountered the issue, so it had been published.

sankhadeeproy007:
This commit was made yesterday, and we haven't published since.

mribbons:
I'm referring to this commit, which I think was published: https://github.com/GeekyAnts/NativeBase/commit/baa88747c45ab469682617e28191e2faa15c6342

PR #551, opened 2017-02-22
Complete customization of Picker button

StevePotter:
The stock Picker didn't allow me to sufficiently customize the button that gets rendered in the form.  This PR lets you customize the rendering of it with a property called renderButton.

Ex:
`
renderButton(onPress) {
return <Button onPress={onPress}>....
}

render() {
  return <Picker renderButton={this.renderButton} ...
}
`

I'd like to see this pattern more and more in NativeBase.  I've found the components are flexible enough most of the time, but when you really need to customize, they lack the appropriate hooks.  The good news it's easy to do.


sankhadeeproy007:
Thanks!

PR #560, opened 2017-02-23
Picker Header Customizations

StevePotter:
Hi, we had a need to customize the Picker (ios only) header's background, Back button color and text.  This was done with the addition of 3 props:

1. headerStyle - style applied to <Header>
2. headerBackButtonStyle - style applied to <Button> inside <Header>
3. headerBackButtonText - text shown instead of hardcoded 'Back'

These are backwards compatible and optional props.  Included in propTypes.  

Like I did with `renderButton`, how do you feel about a `renderHeader()` prop?  This makes it a bit easier to customize the entire header and isn't as tricky as supporting `headerComponent`.  In fact `headerComponent` could be deprecated with a very simple upgrade path.

### DIFF
--- a/src/basic/Picker.ios.js
+++ b/src/basic/Picker.ios.js
@@ -109,12 +109,12 @@ class PickerNB extends Component {
   }
 
   renderHeader() {
-    return (this.props.headerComponent) ? this.modifyHeader() : (<Header >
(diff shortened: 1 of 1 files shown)


basequatro:
Can we merge this one?

SupriyaKalghatgi:

PR #564, opened 2017-02-24
Update typescript definitions

stonecold123:
The typing definitions of React can no longer be installed as a global module. It throws the `cannot find name space React`, so I think it'd be better to add `import * as React from react` at the top.

I also added the definition of `<Item/>`  and updated `<ListItem/>`,`<Icon/>`


stonecold123:
The module name was incorrect in the first file. I'll resubmit later

PR #607, opened 2017-03-06
add Form, Label class and interface

taljacobson:
doesn't look like the definition file was updated to support new Components in 2.0, hope it helps


SupriyaKalghatgi:

PR #674, opened 2017-03-22
Patch index.d.ts to working with typescript

joesonw:
In `index.d.ts`, `react-native` should be used for `ReactNative` specified properties.


aaronpowell:
Yes please! I've just spent the better part of a day trying to 'hack around' it myself. Interim solution is that I'll maintain this file within my own repo and have a step in CI that deletes the bundled one.

aaronpowell:
🎉

PR #686, opened 2017-03-27
Fix BackButton Toast closing

danilvalov:
When we press Android hardware Back Button we see an alert "Modal has been closed". I changed this alert to closing of the Toast.


sankhadeeproy007:
@danilvalov Thanks! :)

PR #729, opened 2017-04-10
Fixed Spinner and Checkbox custom theme integration

stochris:
Fixes the spinner always defaulting to the standard ( local ) theme, not taking into account the context provided theme.
Also fixes the same problem with the checkbox `border` and `backgroundColor`


SupriyaKalghatgi:

stochris:
We need to undo this. Sorry, but a few days I messed up the branches and changed how radio work ( it uses the same icon on ios and android). I'l get back with a new pr in few days. Sorry

PR #758, opened 2017-04-20
[TS] Update definitions

pierre-H:
- Add Toast
- Add Form


pierre-H:
Note : interfaces should have `;` and not `,` according to the ts documentation

pierre-H:
Better to use https://github.com/GeekyAnts/NativeBase/pull/767

PR #804, opened 2017-04-30
Fix the mispelled variable from this.root to this._root

kchalermsook:
In order to link back from List to ListView, the variable name should be this._root.


SupriyaKalghatgi:

PR #808, opened 2017-05-02
Add uppercase prop to Text, make button text uppercase by default on Android

rockmacaca:
This reimplements the behavior available in previous versions of NB. I chose `uppercase` over `capitalize` since it better corresponds to the `text-transform` CSS property value.

To force lowercase text in buttons on Android, the component is used like
```jsx
<Button>
  <Text uppercase={false}>Tap here</Text>
</Button>
```

This also includes some updated TypeScript definitions.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,8 @@ declare module 'native-base' {
     namespace NativeBase {
 
         interface Text extends ReactNative.TextProperties {
-            note?: boolean
+            note?: boolean,
+            uppercase?: boolean
         }
 
         interface Switch extends ReactNative.SwitchProperties { }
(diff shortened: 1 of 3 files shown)


rockmacaca:
@shivrajkumar I noticed in your commit that merges is inconsistent about using `uppercase` or `capitalize` in the propTypes vs what is used in the actual component. The `Text` type definition is also missing the new property. I'll just rebase this PR.

Btw, I really think it should be `uppercase` instead of `capitalize` (though I accidentally said the wrong thing in my commit message). Capitalize Only Affects First Letters while UPPERCASE LOOKS LIKE THIS

rockmacaca:
Ok I rebased and adjusted the commit to only include the relevant stuff. Can you make sure you include the updated type definitions? I need those :P

rockmacaca:
@shivrajkumar sorry to ping you again, but could I at least have some acknowledgement of when this can get integrated? Or what problems you might have with my additions?

SupriyaKalghatgi:

PR #907, opened 2017-05-31
Improve Android button appearances (closes #849, #896)

### LINKED ISSUE #848: Unable to style the tabs
screen shot 2017-05-23 at 1 38 46 pm

How do I bring the tabs down like the one shown below?

screen shot 2017-05-23 at 1 39 43 pm

### LINKED ISSUE #849: Can't create rounded ripple effect for Button in Header
**react-native**: 0.44.0
**react**: 16.0.0-alpha.6
**native-base**: 2.1.4
## Expected behaviour
Button shouldn't resize by adding `rounded` property and should correctly show ripple effect
## Actual behaviour
Button resizes and ripple effect is "square"
## Steps to reproduce
```
<Header>
  <Left style={{ marginLeft: 4, marginRight: -20 }}>
	<Button rounded transparent>
	  <Icon name='menu' />
	</Button>
  </Left>
  <Body>
	<Title>Imports List</Title>
  </Body>
  <Right />
</Header>
```
## Screenshot
image
## Expected example
image
Tested on Android device.

### LINKED ISSUE #896: Button does not support conditional JSX in children list on Android.
RN 0.44
Native Base 2.1.4
React 16.0.0-alpha-6
## Expected behaviour
Add children component with inline conditional JSX in a Button and expect them to show up properly on Android
## Actual behaviour
Get a backtrace as the code tries to iterate over children and access their type property which does not exist for some of them due to JSX condition evaluation ending up in the list of children as a boolean.
Create a Button with 2 children. One Text not showing up because of an inline JSX condition which equals false and another one that is always showing up.
N/A
Android only as the code in Button.js is trying to capitalise only the Android children.
N/A

rockmacaca:
This implements proper behavior for Material Design flat buttons and icon toggles - these buttons ripple as soon as the touch starts. Conveniently, React Native supports masking these ripples to the parent View's border radius. This also fixes a breakage introduced in #808 that closes #896.

This does _not_ completely fix the ripples for regular buttons with rounded corners. However, once React Native actually supports masking bounded ripples (likely if something like facebook/react-native#6515 is reimplemented) they should be fixed here as well.

This also fixes a bug where if you click on a child of the button, the ripple origin is misplaced (see the gif of the before screen with the `block` button).

I have a [demo repo](github.com/rockmacaca/NBButtonPlayground) if you want to see what I've tested so far.

Before:

Proposed:

I've also brought most the button stylings up to MD specs. The way ripple color is determined is as follows:
* Use `variables.androidRippleColorDark` by default.
* Use the `background` or `androidRippleColor` attributes above everything.
* If the `transparent` prop is `true`, use the `backgroundColor` with alpha 0.26 from the `styles` prop
* If the button is not transparent, use the light ripple color (`variables.androidRippleColor`) if the contrast ratio between `backgroundColor` and rgb(0, 0, 0) is less than 3:1.

I still need to test that this doesn't dramatically impact layouts and styles on iOS. I'll update with that here soon.


rockmacaca:
Ok, I just made a few adjustments and now iOS looks pretty good.

iOS before:
screen shot 2017-05-31 at 1 58 21 pm

after:
screen shot 2017-05-31 at 5 43 16 pm

shivrajkumar:
@rockmacaca Looks great, but this breaks some of the others components. Like Bordered/Outline Button, left padding of Left Header Button seems more.

shivrajkumar:
Do check you patch with demo app on NativeBase https://github.com/GeekyAnts/NativeBase-KitchenSink

shivrajkumar:
Closing this due to breaking changes.
You can resolve the above mentioned issues and submit a new PR

ludalex:
This is great. I had to remove rippleEffect since with the current implementation the button press feedback feels sluggish. Is there a chance you can reopen the PR with the current master?

SupriyaKalghatgi:
@ludalex You can ask @specialkk to do so

specialkk:
@ludalex I haven't been very active with the project I was using this for recently, there are also just too many adjustments that have to be done that are potentially breaking changes so I probably won't be doing it any time soon.

Btw, the reason it might feel sluggish is because Android has two separate ripple types: solid buttons show a highlight and then ripple on release, whereas transparent buttons use "borderless" ripples that ripple on the initial touch. Both usually feel ok as long as the highlighted color is clearly visible.

PR #938, opened 2017-06-12
Extra Control for Item inputProps

bahadircivelek:
if this.inputProps not exist program will break.


shivrajkumar:
Merged Similar PR #967

PR #955, opened 2017-06-15
[TS] Added StyleProvider to TypeScript type def

AleksandarDev:


AleksandarDev:
@shivrajkumar Can this be reviewed?

PR #966, opened 2017-06-19
Updates `DeckSwiper` state when receiving new dataSource

mu29:
## Problem

`DeckSwiper` did not update even though I put a new dataSource in it.

## Solution

`componentWillReceiveProps` detects the length of a new data source and changes its state.


mu29:
Conflict fixed! @shivrajkumar

shivrajkumar:
@mu29 Could you also add use-case or code snippet for this

mu29:
Sure.

```jsx
class TestComponent extends Component {

  ...

  render() {
    <Container>
      <Body>
        <DeckSwiper
          dataSource={ this.state.bookmarks }
          renderItem={ quiz => <Image source={ quiz.src } /> }
        />
      </Body>
      <Footer>
        <FooterTab>
          <Button full onPress={ this.onDeleteBookmark } />
        </FooterTab>
      </Footer>
    </Container>
  )
}
```

In earlier versions, calling `onDeleteBookmark()` on these components did not refresh `DeckSwiper`, so the deleted bookmark were remained on the top.

PR #978, opened 2017-06-23
Update README.md

leftynaut:
update readme due to dead link


leftynaut:
Or is [this](https://github.com/GeekyAnts/NativeBase-KitchenSink/blob/master/js/setup.js) the intended link?

PR #1008, opened 2017-07-01
Text Component comma bug fixed

hyunjun19:
react: 
react-native: 

# code
```
const price = 100;
return (<Text>${price}</Text>);
```

output(iOS)
```
$100
```

output(Android)
```
$,100
```

https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/Text.js#L15
children => ['$', 100]

i use ``_.join(children, '')`` explicit join and prevent comma in android


hyunjun19:
sorry. I have bug.
I'll fix and resend.

PR #1149, opened 2017-08-08
CRNA does not require `react-native link`

w-:
Prior to this change, it is confusing/unclear that a user using CRNA or Expo does not need to run `react-native link`.


SupriyaKalghatgi:
@w- Thank you bringing this confusion into our notice.

The ReadMe is updated accordingly.
https://github.com/GeekyAnts/NativeBase/blob/master/README.md#4-getting-started

Issue #640, opened 2017-03-13
Toast callback


justfede:
Hey guys! I wan't to perform some action when the toast closes. It's posible?

Thanks!

shivrajkumar:
@justfede It is not possible do that yet.

alexrussell:
FWIW this is now available, see https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/ToastContainer.js#L79

However, the toast closing code is implemented somewhat non-DRYly, and as such this only actually fires when the toast is *manually* closed. If the original config specified a duration, the toast would still be closed but the callback wouldn't be called as a different code path is run in this case (specifically https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/ToastContainer.js#L59).

SupriyaKalghatgi:
@alexrussell Good one!
@justfede did you take a look?

SupriyaKalghatgi:
Merged #1191

PR #1191, opened 2017-08-26
Add onClose callback to Toast

TimBroddin:
In an app I'm creating I need to navigate somewhere when the button of a Toast is clicked. So I made a callback. Hope you can include this in a next version of NativeBase.


conroywhitney:
Very excited for this RE: #640 and [Docs#24](https://github.com/GeekyAnts/native-base-docs/issues/24) ! 🎉

Issue #1173, opened 2017-08-18
Text is not displayed correctly on Inputs with stackedLabel


Papipo:
## react-native, react and native-base version
react@16.0.0-alpha.12
react-native@0.46.4
native-base@2.3.1

## Expected behaviour
I expect text to be displayed completely

## Actual behaviour
The first character is cut

## Steps to reproduce (code snippet or screenshot)
Just try it on the Kitchen Sink app

## Screenshot of emulator/device
![screen](https://user-images.githubusercontent.com/22678/29472999-4baf7b7a-8456-11e7-8e0c-4614553dd88d.png)

## Is the bug present in both ios and android or in any one of them?
Android. I don't have an iOS device to test.

## Any other additional info which would help us debug the issue quicker.
The first char on the screenshot is a "B".
I am using floatingLabel. I think it also happens with stacked ones.

andreimc:
I have a fix on my current master - submitted a PR

ceefour:
I also confirm this is happening, not just with my app but with nativebase's own KitchenSink as of 31 Aug 2017. Android 7.0 on Asus Zenfone 3.

joaozitopolo:
@andreimc there are some scheduled date for update of the native-base package at npm, with the correction?

shivrajkumar:
Fixed in version 2.3.2

PR #1194, opened 2017-08-28
Fix for #1173 no negative margin left for input inside stacked form or…

andreimc:
… floating form


andreimc:
yeah it brings the text 5 px to the right. So I think the default needs to be -5px and the floated and stack 0 for the marginLeft.

PR #1202, opened 2017-08-30
Set layout handler on button when tab has no name

preynal:
This was causing the tabs to not be measured, and breaking the indicator layout on scrollable tabs.


sankhadeeproy007:
@preynal Looks good to me 👍  cc @shivrajkumar

Issue #1256, opened 2017-09-28
How to remove ripples on Android?


boringer:
## react-native, react and native-base version

react-native: 0.48
react: 16.0.0-alpha.12
native-base: 2.3.2

## Expected behaviour

Remove ripples anywhere on Android.

## Actual behaviour

Still exist.

## Steps to reproduce (code snippet or screenshot)

Wrap my root element with `<StyleProvider style={getTheme(commonColor)}>`, change `androidRipple` in `native-base-theme/variables/commonColor.js` to `false`.

shivrajkumar:
@boringer Fixed in version 2.3.3

PR #1257, opened 2017-09-28
Added missed contextTypes block to Button.js

akhil-geekyants:


akhil-geekyants:
associated with issue #1256

PR #1270, opened 2017-10-03
Make the compile and transpile scripts run on Windows.

AaronFriel:
Adds `rimraf` and `ncp` and replaces *nix utilities in `package.json` `scripts` with Node utilities.

The `dist` folder is not included in these commits because, unfortunately due to differences in path separators, the `dist/` folder has to, be compiled on a posix platform. Otherwise the path separators will be `\` in the generated files.


SupriyaKalghatgi:
@AaronFriel Please look into this
Your PR has failed some of the test cases

PR #1272, opened 2017-10-04
Fix Text component for non-string array children.

VahidBo:
# Motivation
The `Text` component has a bug in case of containing an array and some of its children are not string. For example in case of code below the "Hello" word must not be shown but it's showing before fixing the bug.
`<Text>{true && 'bye bye'} {'33'} {false && 'Hello'}</Text>`

# Test Plan
For this code:
 ```
        <Text>${a}$</Text>
        <Text>{false && 'Hello'}</Text>
        <Text>{false && false}</Text>
        <Text>{true && 'bye bye'} {33} {false && 'Hello'}</Text>
        <Text uppercase>{'bye bye'}</Text>
        <Text uppercase>{true && 'bye bye'} {"333"} {false && 'Hello'}</Text>
        <Text uppercase>{true && 'bye bye'} {33} {false && 'Hello'}</Text>
        <Text><Text>${a}$</Text><Text>${a}$</Text></Text>
```

Before fix:
image

After fix:
image


shivrajkumar:
@VahidBo Amazing Test plan with the PR

PR #1284, opened 2017-10-06
DeckSwiper Swipe Up and fixes

LiewJunTung:
Enable DeckSwiper to do swipe up, plus trigger onSwipeLeft(item) when using swipeLeft() and  onSwipeRight(item) when using swipeRight()


HamadaBrest:
@LiewJunTung I want to block swiping using `onSwipeLeft/Right` ... is this included in your fix ?

shivrajkumar:
@LiewJunTung This is a very specific addition to the deckswiper behaviour. We can still look into its addition if you update the PR.

PR #1290, opened 2017-10-09
Fix mixed ES6 imports with CommonJS exports.

AaronFriel:
When using Webpack/Rollup with ES6 modules, mixed module formats can cause errors.

This fixes that.

Also, I am a bit confused by the file `src/Utils/mapPropsToStyleNames.js`. It appears the authors intended it to be something like a default parameter? But assigning to `mapPropsToStyleNames` without a declaration is an error.

### DIFF
--- a/src/Utils/mapPropsToStyleNames.js
+++ b/src/Utils/mapPropsToStyleNames.js
@@ -1,6 +1,6 @@
 import _ from "lodash";
 
-export default (mapPropsToStyleNames = (styleNames, props) => {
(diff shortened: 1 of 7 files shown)


shivrajkumar:
@AaronFriel The default exports seems to cause some error
simulator screen shot - iphone 6 - 2017-10-13 at 11 47 14
Will look into this and add the changes in the next release

AaronFriel:
Is there a repository you used as a base to find that redbox?

Issue #389, opened 2016-12-27
How to use font-awesome icons or other type of icons that are included?


jsantana90:
I'm having trouble using all of the icons included.

I can only use ionicons or octicons, but if I try to use, for example, [cutlery](http://fontawesome.io/icon/cutlery/) then a `?` symbol appears instead.

sankhadeeproy007:
Change your theme file to set `FontAwesome` as the default Icon font-family.
![screen shot 2016-12-28 at 12 31 54 pm](https://cloud.githubusercontent.com/assets/10998047/21515919/af56c116-ccf9-11e6-9a2f-7b21a1e1af82.png)
Please keep in mind to import the theme file and use it as shown in the screenshot.

OogieBoogieInJSON:
Which was the reason of not setting the icon family using a property on the ```Icon``` component ?

godoakbrutal:
@sankhadeeproy007 
If I don't want to change at the `mainThemeFile`. So how to change iconFamily at `<Icon />` component ?

OogieBoogieInJSON:
@godoakbrutal At this point it's not possible but there is an open merge request about this here: #1319

waritthornnudee:
`<Icon name="flash-off" type="MaterialIcons" />`

type depend on Bundled Icon Sets

Bundled Icon Sets
Entypo
EvilIcons
Feather
FontAwesome
Foundation
Ionicons
MaterialIcons
MaterialCommunityIcons
Octicons
Zocial
SimpleLineIcons

PR #1319, opened 2017-10-21
Accept Icon Type as a prop

yeknava:
In some situations I need to have different icon type for specific icons.


alexrussell:
This would be super useful as it means we can use icons from other sets without having to manually import from `vector-icons` and lose the NB theme support.

rabin111:
I need this! much more easier...

SupriyaKalghatgi:
Related issues: #389

src/basic/IconNB.js
```diff
@@ -65,6 +66,7 @@ class IconNB extends Component {
 
 IconNB.propTypes = {
 	style: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
+	type: PropTypes.string.oneOf(["Ionicons", "Entypo", "FontAwesome", "Foundation", "MaterialIcons", "MaterialCommunityIcons", "Octicons", "Zocial", "SimpleLineIcons"]),
```
alexrussell: Does `PropTypes.string.oneOf([])` work? I'd have though `PropTypes.oneOf([])`

yeknava:
@alexrussell thank you, i fixed that. i thought all rules can be chain but it looks like its specific to isRequired.

alexrussell:
np, just hoping we can get this merged :)

SupriyaKalghatgi:
@yeknava Can you share example snippet with output for this

alexrussell:
Sample intended use:

```jsx
<View>
    <Icon name='person' />
    <Icon type='FontAwesome' name='user' />
</View

//  Will actually 'translate' to the followoing:

<View>
    <Ionicons name='person' />
    <FontAwesome name='user' />
</View>

// (Assuming the icon theme is the default of Ionicons.)
```

The only two issues I see right now is that the entirety of `props` is proxied to the *actual* icon that'll be used, so actually my example above isn't quite correct, as the second example (Font Awesome) will also get the `type` prop passed along. Arguably, this should be removed from the props being proxied.

The second issue is that the `type` prop will only be used if the Icon has a theme set in the context. Now, I think a theme is passed in by default anyway so it's unlikely that there won't be one, but arguably the `type` prop should always set the icon type regardless of the context theme.

alexrussell:
Also, setting `this.Icon` in `cWM` might be asking for trouble - if the `Icon` received a different type in props, the new type will not be used as the previous `Icon` instance won't be unmounted and remounted, but just re-rendered with the new props sent through `cWRP`.

Of course this isn't an issue with this PR per se, it's just the way that this component was originally set up (setting a class property in `cWM`), but this PR makes it meaningful that this is done as a class property in `cWM` so it is something that should be addressed by this PR.

That said, I'd say that for now the PR is fine to be merged, and the things I've mentioned can be documented as current caveats, which can be fixed later.

yeknava:
@alexrussell i like your approach but i see this problem in any NB components. they lacks of customization when i compare them with [RNE](https://react-native-training.github.io/react-native-elements/) or other react or vuejs components. there are a lot of options as props and accept style also for more customization.

alexrussell:
@yeknava sorry I don't quite understand - isn't this PR specifically to add customisation to the `Icon` component?

All I was saying above, for clarity, is that there's a potential issue with the `type` prop not being respected if it were to change dynamically during runtime. Consider the following code:

```jsx
render () {
  return (
    <View>
      <Icon type={this.state.iconTypeFA ? 'FontAwesome' : 'Ionicons'} name='arrow-up' />
      <Button onPress={() => this.setState({ iconTypeFA: ! Boolean(this.state.iconTypeFA) })}>
        Change icon type
      </Button>
    </View>
  )
}
```

That will render an "arrow-up" icon in Ionicons and a button underneath which switches the icon's type to 'FontAwesome'. But using the current code as in this PR, the icon's type will not update to FontAwesome as the `Icon` itself is not unmounted and remounted, but instead just sent new props and asked to re-render.

yeknava:
@alexrussell nice catch, i addressed dynamic type change issue in patch-2 (didnt tested it yet btw)

alexrussell:
Looks like it'd probably do the job, though I'd be inclined to actually go back to the original code that did the theme decision on mount and just do the props check in the `render` method - if `props.type` exists use that, otherwise use the `this.Icon` value we worked out before.

yeknava:
@alexrussell its still get icon type from theme but in some cases we need to overwrite theme icon type for a specific icon. (i had a case like this)

alexrussell:
Not sure what you're getting at there. No worries anyway, it's all good.

OogieBoogieInJSON:
@alexrussell Some set of icons might not have the icon you need but the other one has it so we should be able to decide which icon set to use at component level.

alexrussell:
@OogieBoogieInJSON I totally get that (I even gave the example of use that @SupriyaKalghatgi asked for), but that's what this PR does. I'm all for that.

webdevsyd:
@alexrussell is this already implemented, i'm currently using native base 2.3.3

SupriyaKalghatgi:
@webdevsyd this is not merged yet, that's why this PR is still open

SupriyaKalghatgi:
@yeknava Please check
simulator screen shot - iphone 7 - 2018-01-12 at 15 19 40

alexrussell:
@SupriyaKalghatgi @yeknava yeah looks like line 61 needs to be `this.setIcon` not just `setIcon`.

yeknava:
@alexrussell thanks, it happens when changing code on github's text editor :smile:

JeremyBradshaw7:
Works but getting this error in typescript:
image

SupriyaKalghatgi:
@JeremyBradshaw7 #1594

JeremyBradshaw7:
Thanks @SupriyaKalghatgi - this will hopefully be merged soon.

PR #1304, opened 2017-10-15
add hide method to Toast

LonelyGriffin:
I added method to Toast for close it manually. But since Toast has one instance at a time, I did only one method for this
[issue: #1287](https://github.com/GeekyAnts/NativeBase/issues/1287)

PR #1323, opened 2017-10-24
DRY up the ToastContainer so that there is only one method to close the toast.

alexrussell:
Previously the `closeToast()` method was being called only when the button was pressed. In the case of a duration timeout, a custom bit of code was being run. However, the code was the same code (in terms of hiding the toast) but lacking the `onClose` callback that was added in #1191. Having one place where the toast is hidden will unify the experience here.


LonelyGriffin:
Your DRY is superfluous, because it's done in my PR #1304
and there is one more thing:
onClose callback will called as a result of calling the method cloaseToast by timer
```js
showToast({ config }) {
		this.setState({
                  ...
		});
		if (config.duration > 0) {
			setTimeout(() => { this.closeToast(); }, config.duration);
		}
...
closeToast() {
		const { onClose } = this.state;

		if(onClose && typeof onClose === "function") {
			onClose();
		}

		Animated.timing(this.state.fadeAnim, {
			toValue: 0,
			duration: 200,
		}).start();
		setTimeout(() => {
			this.setState({
				modalVisible: false,
			});
		}, 500);
	}
```
this was not before, it can break back compatibility, because it changes the behavior of onClose callback

alexrussell:
It changes the behaviour on purpose - currently the "onClose" callback is not being called when the toast naturally closes with a timeout. This seems to me to be incorrect behaviour, so this PR is actually fixing that *by* DRYing the code up.

As for conflicts, yes this is true, but I guess whichever of these is merged first will define the baseline code for the other to rebase their PR on.

SupriyaKalghatgi:
@alexrussell Can I know the reason for closing this?

alexrussell:
As mentioned in #1542 and #1324, I accidentally deleted my source branch. But also, this was a refactor that diverged from the latest work on master where work has been done on the toast container. I still stand by my wish to DRY this up *and* make it more consistent, but I didn't see it going anywhere as these PRs are just being ignored anyway. I'm more than happy to have another shot at this PR and post as a new one though like I did with #1324 -> #1542. But it'd be nice to know I'm not wasting my time and that it would be looked at.

SupriyaKalghatgi:
@alexrussell We appreciate your contribution.
There may be delay at times, inconvenience is regretted.
Please go ahead and send PRs

PR #1337, opened 2017-10-31
Fix initialPage prop of Tabs component for IOS

VahidBo:
# Motivation

The `initialPage` prop of `Tabs` component is not working in IOS devices.

# Test Plan

For this code:
```
...
        <Tabs initialPage={2}>
          <Tab heading="Tab1">
            <Tab1 />
          </Tab>
          <Tab heading="Tab2">
            <Tab2 />
          </Tab>
          <Tab heading="Tab3">
            <Tab3 />
          </Tab>
        </Tabs>
...
```
Before fixing this issue, the content of the initial tab (in this case `Tab3`) will render but the first tab is showing and then nothing will be showing in the tab. After this fixing, the right tab will show in the page. In this way the `initialPage` prop will work properly in IOS devices.


akhil-geekyants:
@VahidBo tried your fix. Don't think the issue is fixed.

Gif

tabs

tiberiumihai:
This is because InteractionManager.runAfterInteractions(f) happen to sometimes never call f.

I've added a PR here: #1376

SupriyaKalghatgi:
@VahidBo Are you planning to fix this PR?

tiberiumihai:
@SupriyaKalghatgi I fixed this in #1376. It's basically the same fix as @VahidBo provided, only that is included a fix for IteractionManager which happens to never call the callback sent to runAfterInteractions.

And because of this, ```InteractionManager.runAfterInteractions(scrollFn);``` never run.

You can check the whole fix in #1376.

PR #1366, opened 2017-11-12
add ZIndex to Header's view

wsmlby:
Header's Zindex is not settable from application and it can be covered under, which is almost never what we want. 

give it a huge number.


wsmlby:
https://github.com/GeekyAnts/NativeBase/issues/1367

wsmlby:
This is not ideal, we should provide this as an option instead of hardcode. Please advise.


Also, seems I should not check in the dist/ part, please advise on what to do with it.
Thanks.

PR #1376, opened 2017-11-17
Fix initialPage for Tabs

tiberiumihai:
InteractionManager.runAfterInteractions(f) happen to sometimes never call f. Specially in DEV mode.

For more info check: [https://github.com/facebook/react-native/issues/8624#issuecomment-231040370](https://github.com/facebook/react-native/issues/8624)


SupriyaKalghatgi:
@tiberiumihai Seems like you did not go through Contribution Guidelines

Please check it and update your code

SupriyaKalghatgi:
You cannot edit code in `dist`

tiberiumihai:
My bad. I'll update the PR.

SupriyaKalghatgi:
@tiberiumihai Any updates on this

tiberiumihai:
I used the contributing guideline to also test this.

This should be solved now.

Gif Demo - NativeBase: Fix Initial Tab

ecnaidar:
@tiberiumihai 
Fix works in simple test case (still incorrectly renders in debug mode, though).
In the actual application does not work for both debug and release

Issue is rather frustrating and apparently somewhere deep in RN, since I was able to follow call to scrollTo in Interaction Manager. Correct values are going through up to the place where it is bridged to native module aaand ✨magically✨ nothing happens.

tiberiumihai:
What do you mean "in simple test case"?
Do you have another use case in which this fix doesn't work?

Try to increase the InteractionManager's setTimeout time, probably your app is not so responsive due to some js processing and it needs more time to run the function.

ecnaidar:
@tiberiumihai Just several tabs with text.
I have tried fiddling with timeout time, did not help. I suspect that animations in higher component hierarchy are interfering.

ecnaidar:
#1443 This fixes issue for my specific case so apparently it makes situation better but certainly not fixes the issue

SupriyaKalghatgi:
@tiberiumihai can you check from branch `development` if `initialPage` works as expected?

SupriyaKalghatgi:
@tiberiumihai Indicator issue in case of changing `initialPage` value while hot reload

aliwesome:
this issue is not fixed yet!

Issue #1371, opened 2017-11-14
Swipe treshold for the Tabs component


mahomahoxd:
Hello,

I'm happy with using the native-base elements, there is one thing I wanted to ask though. Is there a way to use the tabs in such a way with a list inside it? I have 2 tabs, in each I have a list. However on refreshing it will freeze the refreshing icon if you swipe just a tiny bit left or right (accidentally even). Isn't there a workaround or a property I can set such as some kind of a swipe threshold? Other than that everything seems great.

Thanks in advance!

glodfinch:
I also came across this issue today. It seems to be the same as being described here #1168 
I tried applying the RefreshControl to a Content that was a parent to the Tabs component and this did work, but it scrolls the page so that the Tabs component disappears. It seems adjusting the swipe threshold would be the solution.

mahomahoxd:
I think if you put tabs in a component such as Content, it will have that kind of behavior. I assume Tabs is a wrapper around a View, this would mean you can just put the RefreshControl thing inside the tabs (I think). I just wanted a swipe threshold as it seems a lot easier to implement that rather than a RefreshControl in each of my lists.

glodfinch:
It's a wrapper around a ScrollView, it uses that element's horizontal scrolling under the hood. Seems it doesn't pass down the refreshControl prop however, but I edited it to do that and it seems to work fine! Here, after line 176:
https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/Tabs/index.js

I added:
`refreshControl: this.props.refreshControl,`

And it passes down the prop and works perfectly from my brief testing.

mahomahoxd:
You could make a pull request, I'd also put it in the proptypes between line 30 and 42, but thats a minor thing. I'll test that out too.

mahomahoxd:
I've tested it as well, it seems to work fluently

glodfinch:
Found some issues with this - closed PR, see my comment for info

SupriyaKalghatgi:
#38

Issue #1168, opened 2017-08-16
refreshControl in tab


Martin25699:
Hello, after add package (zo0r/react-native-push-notification or facebook/react-native-fbsdk)  loading wheel hangs on the swipe
Version
```
"dependencies": {
		"native-base": "^2.3.1",
		"react": "16.0.0-alpha.12",
		"react-native": "0.47.1",
		"react-native-push-notification": "^3.0.0"
	}
```
Code page

```
import React, { Component } from 'react';
import { AppRegistry, RefreshControl } from 'react-native'
import { Container, Header, Content, Tab, Tabs } from 'native-base';
export default class AwesomeNativeBase extends Component {
    constructor(props) {
        super(props);
        this.state = {
            refreshing: false
        };
    }
    updateContent() {
        this.setState({refreshing:true});
        setTimeout(()=>{
            this.setState({refreshing:false});
        },2000);
    }
    render() {
        return (
            <Container>
                <Header hasTabs />
                <Tabs initialPage={1}>
                    <Tab heading="Tab1">
                        <Content refreshControl={<RefreshControl
                            refreshing={this.state.refreshing}
                            onRefresh={()=>{this.updateContent()}}/>
                        }>
                        </Content>
                    </Tab>
                    <Tab heading="Tab2">
                        <Content refreshControl={<RefreshControl
                            refreshing={this.state.refreshing}
                            onRefresh={()=>{this.updateContent()}}/>
                        }>
                        </Content>
                    </Tab>
                    <Tab heading="Tab3">
                        <Content refreshControl={<RefreshControl
                            refreshing={this.state.refreshing}
                            onRefresh={()=>{this.updateContent()}}/>
                        }>
                        </Content>
                    </Tab>
                </Tabs>
            </Container>
        );
    }
}
AppRegistry.registerComponent('AwesomeNativeBase', () => AwesomeNativeBase);
```
![tab_bah](https://user-images.githubusercontent.com/10417137/29362446-7d0a6a98-82a5-11e7-8ef3-0dad07e6fa20.gif)

sitompul:
It's because you drag the tab into another tab while pulling the refresh trigger down. My workaround is to disable the sliding part or if you insist of doing the sliding refresh, you must check **react-native-tab-view**.

shivrajkumar:
[...]

PR #1384, opened 2017-11-21
Pass down refreshControl prop to ScrollView

glodfinch:
This fixes some issue with RefreshControl usage on the Tabs element as per #1371 and #1168


glodfinch:
After some further use I've noticed the refresh control prevents scrolling up in the tabs. I couldn't find a way to avoid that. I also noticed in the react docs:
> Only works for vertical ScrollViews (horizontal prop must be false).
horizontal prop is true in this case, so I'm not sure why it worked at all, but that means it's hacky and likely a bad idea to use it at all (not to mention it doesn't work correctly.)
Thus I'm closing the pull request.

Issue #1102, opened 2017-07-25
Floating Label with Error


kelvinblade:
I am trying to display the red underline error with the red cross icon shown here https://docs.nativebase.io/Components.html#error-textbox-headref with the  floating Label shown here https://docs.nativebase.io/Components.html#floating-label-headref.

`<Item floatingLabel>
  <Label>Username</Label>
              <Input />
            </Item>`

How do I add the error prop "error" to Item conditionally? `<Item error>`

Thanks.

`<Item floatingLabel error>
          <Label>{field.placeholder}</Label>
          <Input
            className="form-control"
            type={field.type}
            {...input}
            disabled={disabled}
            />
            {touched ? <Icon name='close-circle' style={styles.error} /> : <Text />}
        </Item>`

I tried this but the red line is always showing and the icon is on the left side of the text

shivrajkumar:
@kelvinblade 
```
<Item floatingLabel error>
 <Label>Username</Label>
 <Input />
 <Icon name="close" />
</Item>
```
should work.
But Icons with FloatingLabel shows only on left.
We are working on to fix it.

kelvinblade:
@shivrajkumar 

thanks for the reply, it works but how can i only show the red line when there's an error?

ardifirmansyah8:
is this bug fix? i need floating label with icon on right of input. I want to create password field with show / hide password masking.

Garrett-Smith-iq:
Until this issue is closed you can put the icon on the right with style.

```
<Item style={{ flexDirection: 'row-reverse' }} >
```

akhil-geekyants:
Made a PR #1356

TomMahle:
Is there a good way to display error text in a floatingLabel item? 
This does not display the text at all, and was all I could find in the documentation.

        <Item error={hasError} floatingLabel={true}>
            {hasError ? <Text>{error}</Text> : <Text />}
            <Label>{label}</Label>
            <Input {...passThroughInputProps} {...passThroughProps} />
        </Item>

chinmay-xenious:
This issue is not solved by the style={{ flexDirection: 'row-reverse' }},
Because after use this style, item left margin was swap to right.

Any more option?

chinmay-xenious:
Also there is an icon position issue for floatingLabel Item with 'last' property.

ettaibi:
having same issue, Icons on the right with floating label doesn't work

SupriyaKalghatgi:
[...]

PR #1406, opened 2017-11-29
Fix for issue #1102.

akhil-geekyants:
Issue - Floating Label with Error breaks UI


SupriyaKalghatgi:
@akhil-geekyants Many cases not considered with this PR
Can you reconsider this

PR #1435, opened 2017-12-09
Fix crash when scrolling swiped list with removed item

marano:
How to reproduce issue:
* Render a list with the swipe gesture
* Swipe one list item
* Dynamically remove the item from the list
* Scroll list

Result:
Excpetion is thrown.

Expected result:
No error.

It looks like the NativeBase swipe feature was ported from jemise111/react-native-swipe-list-view . You can find an issue there for the exact same problem at https://github.com/jemise111/react-native-swipe-list-view/issues/9, and the fix at https://github.com/jemise111/react-native-swipe-list-view/pull/15


SupriyaKalghatgi:
@marano Can you elaborate the purpose of this PR along with use case ?

marano:
Hey @SupriyaKalghatgi! Thank you for the heads up, I just added some info.

SupriyaKalghatgi:
@marano I dont find this issue with NativeBase 2.3.7
Can you check it

marano:
Hi @SupriyaKalghatgi, yes I can confirm it is still an issue. You can check it on the video below

ffgw3usgix

SupriyaKalghatgi:
@marano Oh ok!
Can you share your code snippet for the same?

marano:
@SupriyaKalghatgi yes sure! you can find this specific screen code here https://github.com/seedtronic/doctalk-mobile/blob/master/components/DoctorAgenda.js

SupriyaKalghatgi:
@marano Looks like i need to install many things
It would be easy for me to check this on snack

marano:
@SupriyaKalghatgi I am trying to reproduce it here https://snack.expo.io/BJYVnH18G

But the `closeOnRowBeginSwipe` flag is not working, which means I can't get to the state where the issue occurs. Do you know if it is a bug with the `closeOnRowBeginSwipe` option?

SupriyaKalghatgi:
@marano whatever is the snippet available on the provided snack link, worked as expected

marano:
@SupriyaKalghatgi No it isn't, the `closeOnRowBeginSwipe` is not auto closing an opened row when I swipe the list.

The bug this PR only happens when the list item is auto closed when scrolling, if the auto close doesn't work the error won't occur. That's why I am having problems reproducing the error on the snack, because there seems to be an issue with `closeOnRowBeginSwipe`. Does it make sense?

SupriyaKalghatgi:
@marano Ok!
In that case, can you make this code snippet available on new branch in your repo?
With required dependencies only

SupriyaKalghatgi:
@marano This PR will be closed in a day or two, in case of no activity

PR #1443, opened 2017-12-11
Tabs initial setup change

ecnaidar:
<Tabs /> underline is not properly positioned when there is a lot of stuff to render.
This fix eliminates issue in my case, but simply putting around hundred <Text /> children breaks it 100% on iOS.
Further investigation required

### DIFF
--- a/src/basic/Tabs/index.js
+++ b/src/basic/Tabs/index.js
@@ -66,9 +66,8 @@ const ScrollableTabView = createReactClass({
 
   componentDidMount() {
     const scrollFn = () => {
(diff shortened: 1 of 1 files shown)


SupriyaKalghatgi:
@ecnaidar 
> underline is not properly positioned when there is a lot of stuff to render.
Unable to replicate this. Can you share code snippet and screenshot for same?

SupriyaKalghatgi:
@ecnaidar Indicator issue in case of changing `initialPage` value while hot reload

PR #1445, opened 2017-12-11
rtl support added for scrollable tabbar

ansal-exp:
### BACKGROUND

I know that native-base has no immediate future plan to support RTL, even though most the components in native-base seems RTL ready. During development of an application which uses scrollable tab bar, I got stuck by the RTL issue in the ScrollableTab component. I got some spare time to look into the code and got a solution which will not broke anything and supports RTL out of the box.

### Problem

The problem was when we render the scrollableTab in RTL layouts, it works fine and the only drawback was floating indicator doesn't have any effects. That is, it will be in LTR format.
The problem was it doesnt check for RTL layouts and the floating indicator was rendered using absolute positions, with an animated left (initial value : 0)

### Solution

I added the check for RTL layout and also added a animated right (initial value: 0) and updated it in the same way as the left value. And it works.

##### side note

This is my first pull request to an external library. I tried my level best to write and test the changes. If I missed something, please help me to improve. 
thanks.


ansal-exp:
I am closing this PR since I found some critical issues on Android. I will re-open it once if I manage to successfully tackle those.

SupriyaKalghatgi:
@ansal-exp Sure

AmeerTgh:
is rtl supported for scrollable tabbar in last version?

PR #1341, opened 2017-11-03
 Tab's heading 

wallel:
to fix issue #1301

PR #1449, opened 2017-12-12
(typescript) Add missing tabBarUnderlineStyle property to Tabs 

april-ayres:
Added the tabBarUnderlineStyle property as it was missing from the typescript definitions.


SupriyaKalghatgi:
#1341

Issue #1301, opened 2017-10-13
Type 'Element' is not assignable to type 'TabHeading'


mengyiyuan:
## react-native, react and native-base version
react-native: 0.49.1
react: 16.0.0
native-base: 2.3.2
typescript: 2.5.3
## Expected behaviour
Tab component should accept a heading props of type TabHeading
## Actual behaviour
Typescript complains that:
"Type '{ heading: Element; children: Element; }' is not assignable to type 'Readonly<Tab>'.
    Types of property 'heading' are incompatible.
      Type 'Element' is not assignable to type 'TabHeading'."
## Steps to reproduce (code snippet or screenshot)
```
class TypeTabs extends React.Component {
    public render() {
        return (
            <Header hasTabs={true}>
            <Tabs>
                <Tab heading={"Search By Name"}>
                    <SearchPage />
                </Tab>
                <Tab heading={ <TabHeading><Text>Fuzzy Search</Text></TabHeading>}>
                    <SearchPage />
                </Tab>
          </Tabs>
            </Header>
        );
    }
}
```

## Screenshot of emulator/device
Cannot compile by running `npm run start:ios` due to the error above.
## Is the bug present in both ios and android or in any one of them?
Same, because it's compilation error.
## Any other additional info which would help us debug the issue quicker.
As shown above, I have followed the example both in simple tabs and advanced tabs format, but none works. The documentation I was referring to is here: https://docs.nativebase.io/Components.html#tabs-def-headref

mengyiyuan:
After digging into the index.d.ts file, it seems that the `heading` prop of the `Tab` element is wrongly bind to the `TabHeading` class, instead of a React Native component type. Is there any reason behind this? How do you get it to properly work with TypeScript's type checking.

akhil-geekyants:
@mengyiyuan couldn't reproduce the error. Attaching a screen shot 
![tabheading](https://user-images.githubusercontent.com/27768184/31599559-91d9ca18-b270-11e7-8be1-7b55f0d1b223.png)
Code
```
 <Header hasTabs={true}>
      <Tabs>
         <Tab heading={"Search By Name"}>
              <View/>
          </Tab>
          <Tab heading={ <TabHeading><Text>Fuzzy Search</Text></TabHeading>}>
              <View/>
          </Tab>
       </Tabs>
  </Header>
```
TabHeading is binded with Tab because we recommend using `<TabHeading/>` with header prop of `<Tab/>`

[...]

PR #1465, opened 2017-12-22
fix bug - Type 'Element' is not assignable to type 'TabHeading' #1301

qinyang1980:
fix this bug : 
https://github.com/GeekyAnts/NativeBase/issues/1301

modify the index.d.ts file.


SupriyaKalghatgi:
@qinyang1980 There's a PR for this #1341

PR #1480, opened 2018-01-02
Modified color values in variable files

akhil-geekyants:


SupriyaKalghatgi:
Continuation for #1463

PR #1495, opened 2018-01-09
Update index.d.ts

fanzkday:
There is no left/body/ attribute on SwipeRow when use Typescript.


SupriyaKalghatgi:
@fanzkday Can I know the reason for closing this?

PR #1507, opened 2018-01-12
Made some SwipeRow properties optional

akhil-geekyants:


akhil-geekyants:
This PR is to prevent tslint error, when some SwipeRow properties are not present.

PR #1508, opened 2018-01-12
fix(toast): respect MD spec for padding

realappie:
According to the following image in the material design specification, padding is not 10 pixels on all sides.

material design spec 

See the [material design specification](https://material.io/guidelines/components/snackbars-toasts.html#snackbars-toasts-specs) on snackbars and toasts for more information.


SupriyaKalghatgi:
@realappie Sorry for delay
Material has updated design guidelines, so need more work here

PR #1324, opened 2017-10-24
WIP: Fix toast timeout

alexrussell:
Currently when a toast is closed, its "duraiton" timeout clock is still going. So if you dismiss a toast and then show another before that timeout elapses, the second toast will hide much sooner than its duration.

This PR stores the timeout's ID and simply clears it when a call to `show` is made.

---

WIP: This PR builds upon #1323 so that one should be merged first, and then this one rebased on top, consisting, eventually, of a single commit.

PR #1542, opened 2018-01-26
Fix toast timeout bug

alexrussell:
Save the timeout ID when a toast is shown so that we can clear any existing timeout when a new toast is shown so that an old timeout doesn't close a new toast prematurely.

---

Replaces #1324 as I accidentally deleted my branch when trying to tidy up. Also I have redone the work on top of the latest master.

@jswxwxf you should now follow this issue.


SupriyaKalghatgi:
@alexrussell Thank you for the update!

eassymo:
Another Toast timeout bug is when duration is not defined or is zero (0), Toast should be sticky or permanent but Toast auto-close after shown, otherwise add a property called "sticky" or "closeByButton", anything, but we could not have a sticky toast.

Out of "timeout" toping, Toast should be "swipeable" and call "onClose" event when timeout or swiped, and buttonText should have another action like "onPress" not call onClose.

SupriyaKalghatgi:
@eassymo Toast takes default value of `1500` for duration

alexrussell:
@eassymo You should open up some issues from what you said here as some make sense to me. I think I can see at least three separate issues you can open here:

1. There should be some way to make a permanent (i.e. must be dismissed) toast. Either setting timeout to a special value (`0`, `-1`) or using a specific config of `permanent: true` or `autohide: false`, etc.
2. Toasts to be swipeable to dismiss
3. Separate callbacks for the way that a toast was closed. In order to conserve BC I imagine we need to leave the current `onClose` as happening when the toast closes through click *or* timeout, but introduce two more: `onDismiss` that only gets called if the toast was user-dismissed and `onAutoClose` that only gets called if the toast closed due to the timeout expiring.

ahartzog:
Created a PR for onDismiss --> https://github.com/GeekyAnts/NativeBase/pull/2678/files

PR #1547, opened 2018-01-29
Fixed Floating label cut off in RN 52 fixes #1532

### LINKED ISSUE #1532: Floating label cut off

- React-Native 0.52.0

- React: 16.2.0

- Native-Base: 2.3.7

## Expected behaviour
When using an input with a floating label, the entire label should be visible when floating.
## Actual behaviour
Once the label starts floating, the top half of the label is cut off.
Something as simple as the following basic case causes it:
```
<Form>
  <Item floatingLabel>
    <Label>Test</Label>
    <Input />
  </Item>
  <Item floatingLabel>
    <Label>Test</Label>
    <Input />
  </Item>
  ...
</Form>
```
screenshot_1516514113

I am not able to test it on iOS at the moment.
It used to work a while ago but I only noticed it is broken now, not sure on which version it broke.

SupriyaKalghatgi:


fendorio:
+1

PR #1549, opened 2018-01-29
Add accessibility support for checkbox

IsracardTeam:


SupriyaKalghatgi:
This PR cannot be accepted, since we have not supported it so far with NativeBase
This is something specific and not generic

PR #1556, opened 2018-01-31
Replaced getters with variables

akhil-geekyants:


akhil-geekyants:
This PR is incomplete. Replacing getter functions with variables didn't give values (values were undefined).

PR #1562, opened 2018-02-02
DRY-up ToastContainer

alexrussell:
- Fix broken eslintrc file
- Refactor `ToastContainer` to DRY-up calls to `Animated.timing` by:
    - Removing multiple calls to setTimeout by extracting the conditional to a variable
    - Passing references to `this.closeToast` in the timeout rather than redefining the function's body in the timeout callback
- Save a timeout when fading the toast out by using the `Animated.timing` completion callback rather than using `setTimeout` with a 500ms delay.


SupriyaKalghatgi:
@alexrussell Can you run this snippet of use case, with respect to this PR
```
<Button
  onPress={() =>
    Toast.show({
      text: "Wrong password!",
      buttonText: "Okay",
      duration: 5000
    })}
>
  <Text>Toast</Text>
</Button>
```

PR #1577, opened 2018-02-09
Added icon styles to transparent button in footer, fix #1576

### LINKED ISSUE #1576: Icons in buttons in footer not properly styled
RN: 0.51.0,
React: 16.2.0
NativeBase: 2.3.7

## Expected behaviour
`variables.topTabBarActiveTextColor` changes the color of icons in my footer, including icons that are in buttons.
## Actual behaviour
`variables.topTabBarActiveTextColor` does not change the color of icons in buttons in the footer, or at least not icons in transparent buttons
For a footer that looks like this:
```
<Footer>
  <Right>
    <Button transparent>
       <Icon name="map" ios="ios-globe-outline" android="md-globe" />
    </Button>
  </Right>
<Footer>
```
The icons will always be black, as no styling is attributed to them.
Both
PR: #1577

Shurelia:
Fixes #1576

Icons in buttons in footers are not being styled by the global variables, causing them to always be black. To fix this, I basically copied what was done in Headers.js, to get the global color from commonColor.js to be used as the icon color. Additionally, I added styles for text colors in buttons as well (to match the icon color).

### DIFF
--- a/src/theme/components/Footer.js
+++ b/src/theme/components/Footer.js
@@ -14,7 +14,13 @@ export default (variables = variable) => {
           shadowColor: null,
           shadowOffset: null,
           shadowRadius: null,
-          shadowOpacity: null
+          shadowOpacity: null,
+          "NativeBase.Icon": {
+            color: variables.topTabBarActiveTextColor
+          },
+          "NativeBase.IconNB": {
+            color: variables.topTabBarActiveTextColor
+          }
         },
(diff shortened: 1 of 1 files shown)


Shurelia:
@SupriyaKalghatgi I would like to know how this is invalid. Please look at the surrounding code. What I have submitted is entirely consistent with what is already there.

PR #1586, opened 2018-02-14
Changed 'SubTitle' to 'Subtitle' in index.d.ts

akhil-geekyants:
Fix based on issuecomment
https://github.com/GeekyAnts/NativeBase/pull/1424#issuecomment-365357198


Dome-GER:
Thanks so much for the quick fix!

PR #1596, opened 2018-02-16
Consider Icon type prop when fetching names

fredmon3:
This prevents icon names from being mapped via `NBIcons.json` when prop `type` is configured as something other than `"Ionicons"`.

Without this fix the following name is incorrectly mapped to `"md-send"` which fails to load for type `"MaterialIcons"`:
``` javascript
<Icon name="send" type="MaterialIcons" />
```


fredmon3:
Is there any additional information I can provide to assist with this PR?  Thanks!

PR #1597, opened 2018-02-16
Max height on actionsheet modal

### LINKED ISSUE #1595: ActionSheet on Android exceeds height of screen on long list of items

- react-native: 0.52.2
- react: 16.2.0
- native-base:  2.3.7

## Expected behaviour
When using ActionSheet in Android with a long list of items, the list should be scrollable and bounds of the modal should remain confined within the root container.
## Actual behaviour
The modal height exceeds the height of the container enough that the top of the title is clipped by the top of the content container.
```javascript-jsx
<Button transparent onPress={() => ActionSheet.show(
	{
	   options: collections.map(collection => { return collection.title; }), // Has 12+ entries
	   cancelButtonIndex: 0,
	   title: 'Select Collection'
	},
	buttonIndex => {
		this.changeCollection(collections[buttonIndex]);
	}
)}>
</Button>
```

screen shot 2018-02-15 at 5 45 38 pm

It may appear that the status bar is covering the top, but in reality the container is trimmed to the bottom of the spacebar, otherwise the top of the 'S' and 'C' would be visible under the status bar that has an alpha value on the color.

Android only, ios uses the built in ios actionsheet.

I already found a solution.  Adding `maxHeight: '100%'` to the TouchableOpacity view will give the expected behavior.  I will open a pull request to fix the issue.

a0js:
Resolves #1595. This will make sure that a large modal created for android action sheet stays inside the bounds of the root container.

Before:
screen shot 2018-02-15 at 5 45 38 pm

After:
screen shot 2018-02-15 at 7 20 48 pm

Extra change:  Eslint was unable to read the rc file as it was written. I updated to insure this change would be within the parameters of the eslint guidelines for the project.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,6 @@
 {
   "parser": "babel-eslint",
-  "rules":{
+  {
     "indent": [
       "error",
       2
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,9 +1,11 @@
 {
   "parser": "babel-eslint",
-  "indent": [
-    "error",
-    2
-  ],
+  "rules":{
+    "indent": [
+      "error",
+      2
+    ]
+  },
   "plugins": [
     "react",
     "jsx-a11y",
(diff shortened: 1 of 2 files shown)


a0js:
I removed the eslint fixes as there is already a PR to address the issue.

Issue #1601, opened 2018-02-17
React Native have to double click for onPress to work inside Tab


samjabeen:
Hi! 

Thanks in advance!

I'm using React Native with the Native Base library. Problem is double click needed onPress event to fire on Native Base' Button inside Tab  when the keyboard is open.

Sample code
```
	<Tabs>
		<Tab>
			
			<Content keyboardShouldPersistTaps='always' keyboardDismissMode='on-drag'>
				<Input
					keyboardType = 'numeric'
					keyboardType={Platform.OS === 'ios' ? 'number-pad' : 'numeric'}
                />
				<Button rounded  danger onPress={this.handleUpdateClick}>
                        <Text>Update</Text>
				</Button>
			<Content>
	
		</Tab>
		...
	
	</Tabs>
```
	"react-native": "0.47.1",
	"native-base": "1.1.0"

Expected behaviour
	Need an onPress event to fire on Native Base' Button inside Tab  when the keyboard is open
Actual behaviour
	double click needed for onPress o fire on Native Base' Button inside  Tab when the keyboard is open

SupriyaKalghatgi:
Fixed with NB [2.3.10](https://github.com/GeekyAnts/NativeBase/releases/tag/v2.3.10)

samjabeen:
@SupriyaKalghatgi thanks for the update.

PR #1608, opened 2018-02-19
Fix for issue #1601 (React Native have to double click for onPress to…

akhil-geekyants:
… work inside Tab)


SupriyaKalghatgi:
@akhil-geekyants This PR has not fixed the issue

PR #1621, opened 2018-02-22
add unclickable listitem

future-challenger:
Sometimes people dont need a `ListItem` which supports click.
Espicially in dev mode, there always would be an error, when I scroll the `List` up or down.


SupriyaKalghatgi:
@future-challenger `ListItem` supports click feature when used with `button` prop
Unless that, it behaves normally

Can you elaborate on need of this PR?

SupriyaKalghatgi:
@future-challenger This PR will be closed by end of the week incase of no activity

PR #1630, opened 2018-02-25
Fix style on selected and noborder

lucaskuan:
__Fixed follow style__
1. When use selected style, the BODY text and RIGHT content color doesn't change
2. When use noborder style, it still showing borders


SupriyaKalghatgi:
@lucaskuan `noBorder` works fine, thats not an issue

SupriyaKalghatgi:
#1657

Issue #1640, opened 2018-02-28
Icon type isn't working properly or the Icons are very out-to-date.


rafaelgrilli92:
RN 0.51
NB 2.3.9

I'm just trying to add a "remove" Icon from FontAwesome and I'm getting this:

![image](https://user-images.githubusercontent.com/21044452/36798322-608ab9e2-1cfe-11e8-85c7-e0072837c6d8.png)

```
<Icon 
	type="FontAwesome"
	name="remove" 
/>
```

![image](https://user-images.githubusercontent.com/21044452/36798244-375cb3ae-1cfe-11e8-9dd2-a9154f66b2ac.png)

jarvisluong:
I doublt there is a "type" prop for Icon, as in the docs, this <Icon> component only uses Ionicons from react-native-vector-icons

rafaelgrilli92:
@jarvisluong And why does it comes up with that?
![image](https://user-images.githubusercontent.com/21044452/36828889-89583c9c-1d70-11e8-84de-6a9768691578.png)

jarvisluong:
Oops I was checking the source code again and there is indeed a type prop, but as I saw in the source code, no matter what type of icon we are using, native-base will auto-prepend the icon name with `md-` if we are in android, and `ios-` on ios. That's why FontAwesome can't find md-remove

Source code of Icon.js:
https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/Icon/index.js

rafaelgrilli92:
Yeah, I kinda noticed that on the Icon.js file, but it's weird because it was working like 2 days ago hahaha

Or maybe I was always getting the Ionicons icons and didn't notice it lol

rafaelgrilli92:
Thank you mate! Appreciated :smiley:

SupriyaKalghatgi:
#1596

SupriyaKalghatgi:
Fixed with NB [2.3.10](https://github.com/GeekyAnts/NativeBase/releases/tag/v2.3.10)

PR #1643, opened 2018-03-01
get correct icon name when using custom Icon type

jarvisluong:
when using <Icon> and provide with custom `type` props, we shouldn't auto prefix `md-` and `ios-`
related issue: 
https://github.com/GeekyAnts/NativeBase/issues/1640


SupriyaKalghatgi:
@jarvisluong #1596 initiated prior
Hence closing this

Issue #1274, opened 2017-10-04
List item having margin from left


tundak:
there is bug with `ListItem` having margin from left in Moto g 5.

akhil-geekyants:
Is the margin present only on Moto g5. If it is present in every device you can see this issue #1219

AntonPuko:
also, have unexpected left margin instead of padding in listItem. this doesn't look as expected behavior
![iml1-zxu-km](https://user-images.githubusercontent.com/8849922/31418250-4284dcb2-ae35-11e7-8527-cc9ccf684579.jpg)
phone: galaxy a5.
native-base: 2.3.2

dragonfire1119:
This happens on the iOS Simulator as well. I'm just manually over riding the style with ``marginLeft: 0, paddingLeft: 17`` I'm on native-base: 2.3.2

shivrajkumar:
@dragonfire1119 The margin is actually intended to have some space from the left for both the Text and its bottom border. Using padding doesn't provide that space/gap for the bottom border.
Use List with same background as the ListItem as the wrapper for those ListItems. It will resolve the issue.

AntonPuko:
`Use List with same background as the ListItem as the wrapper for those ListItems. It will resolve the issue.`
Sorry, that I'm a bit rude, but its bad solution. Instead of figuring out how to structure you components right way, you just say: 'make a crunch' and close the issue. ListItem should definitely take full space of list.

dragonfire1119:
@shivrajkumar Thanks for the reply! I understand you on the background & I already knew that. But I want a different background so it contrasts better instead of all white. Which to have a different back you have to use padding. Am I not getting how you designed it? If you designed it to only work with white background why is the background grey?

Ewocker:
Same Issue here, this is definitely not expected behavior, margin on the left is really bugging me a lot.

xyzzyxyzzy0077:
Same thing on iOS, should be a glitch

ashkonhakimi:
I got it to work as follows.  

` width: '100%',
    marginLeft: 0,
    paddingLeft: 0,
    paddingRight: 0,
    marginRight: 0 `

nonameolsson:
Is this something introduced in a newer version? I upgraded, and suddenly I started having this weird leftMargin. `paddingLeft` should be used.

ashkonhakimi:
Strange. I have the newest version I did not run into that.

nonameolsson:
[...]

PR #1661, opened 2018-03-06
Use padding instead of margin in ListItem to fix hover states

cmmartin:


SupriyaKalghatgi:
#1274

SupriyaKalghatgi:
This PR does not fix #1274
Check #1712 for fix

Issue #1397, opened 2017-11-27
Bug Components actionsheet error: null is not an object (evaluationg 'this.actionsheetInstance._root.showActionSheet')


Angelk90:
@GeekyAnts, @akhil-geekyants: Hi, I'm using the ActionSheet component on android.
Link: https://docs.nativebase.io/Components.html#actionsheet-def-headref
Every time I always make the same mistake, see the picture:

I tried using the Root tag and then also the Container, only the Root, even the Container but always the same mistake.

You know how to fix the problem, it's a problem that you always get, every single time.
Once it's done, the next turn stops working.
![errorj](https://user-images.githubusercontent.com/20476002/33278123-d9bc2ae2-d39a-11e7-83d3-019f110df01d.png)

## react-native, react and native-base version
		"native-base": "^2.3.3",
		"react": "16.0.0",
		"react-native": "0.50.4",
## Expected behaviour
See picture.
![errorj](https://raw.githubusercontent.com/GeekyAnts/NativeBase-KitchenSink/v2.2.0/screenshots/android/actionsheet.gif)

## Actual behaviour
As soon as you click the button you will see that error message.
## Steps to reproduce (code snippet or screenshot)
```
import React, { Component } from 'react';
import { Text, StyleSheet, TouchableOpacity } from 'react-native';
import { ActionSheet, Root, Container, Content } from "native-base";  // 2.3.3
var BUTTONS = [
  { text: "a", icon: "play", iconColor: "#2c8ef4" },
  { text: "b", icon: "shuffle", iconColor: "#000" },
  { text: "c", icon: "clipboard", iconColor: "#ea943b" },
  { text: "d", icon: "trash", iconColor: "#fa213b" },
];
var CANCEL_INDEX = 2;

export default class App extends Component {
  render() {
    return (
      <Root>
        <Container>
          <Content>
            <TouchableOpacity
              onPress={() =>
              ActionSheet.show({
              options: BUTTONS,
              cancelButtonIndex: CANCEL_INDEX,
              title: "title"
              },
              buttonIndex => {
                alert(buttonIndex)
              }
              )}
              underlayColor="transparent"
              activeOpacity={1}>
              <Text style={styles.text}>Name</Text>
            </TouchableOpacity>
          </Content>
        </Container>
      </Root>
    );
  }
}

const styles = StyleSheet.create({
  text: {
    fontSize: 18,
    color: '#34495e',
  },
});
```

[...]

Issue #1315, opened 2017-10-20
Actionsheet and Picker are bad (Android)


hoboman313:
Actionsheet:
- largely unconfigurable and unstylable
- works 90% of the time, every time. https://github.com/GeekyAnts/NativeBase/issues/937 was never fixed properly which leads to random crashes.
- It just doesn't look or feel right on Android. I have not seen this type of pattern in any Google nor other apps.

Picker:
- does anyone here think that it's okay to have all that real state, but only make the tiny dropdown clickable?

AbdallaMohamed:
The bottom sheet should slide up instead of fading in. https://material.io/guidelines/components/bottom-sheets.html

akhil-geekyants:
@AbdallaMohamed considering this. Will be implemented in the near future.

jobala:
@akhil-geekyants I can work on this enhancement.

akhil-geekyants:
@jobala once completed, you can make a PR. Let us know if you need any help.

jobala:
I can make this use, the native bottom sheet, should I do that?

SupriyaKalghatgi:
Fixed with [2.5.0](https://github.com/GeekyAnts/NativeBase/tree/v2.5.0)

akhil-geekyants:
- Issue #937 is fixed with native-base version 2.5.0.

- Implementation of ActionSheet on android is an implementation of the following screen 

![screen shot 2018-06-12 at 6 35 02 pm](https://user-images.githubusercontent.com/27768184/41292216-8a3ab32e-6e6f-11e8-82bc-0e3fe40b5b3f.png)

- NativeBase Picker for Android is a direct implementation of React Native Picker.

    Let us know if you would like us to add anything.

SupriyaKalghatgi:
This issue will be closed this week in case of no response

ardyfeb:
> Fixed with [2.5.0](https://github.com/GeekyAnts/NativeBase/tree/v2.5.0)

Let me know how make actionsheet with fade animation,  Can you give some example code ?

rohithramachandran:
I am on version 2.13.0.. Looks like its still fading in instead of slide up... 
`ActionSheet.show()`

Issue #937, opened 2017-06-12
ActionSheet.show crash 2.1.5


fbacker:
Upgraded react-native and nativebase. No changes in my code. Running on iOS.

## versions
    
    react: 16.0.0-alpha.12
    react-native: 0.45.1
    native-base: 2.1.5

## Expected behaviour
ActionSheet.show will show the ActionSheet

## Actual behaviour
App crashes

    Cannot read property 'showActionSheet' of null
    Function.show
    ActionSheet.js:128:85

![image](https://user-images.githubusercontent.com/4478361/27021861-5c3474ea-4f4b-11e7-897d-ca70aba95da6.png)

fbacker:
If I add this on a random view-container it works

    <ActionSheet ref={(c) => { ActionSheet.actionsheetInstance = c; }} />

SupriyaKalghatgi:
![actionsheet](https://user-images.githubusercontent.com/13464885/27166207-cd384f3c-51b6-11e7-9116-bdb43ba8c431.gif)

```
"dependencies": {
    "native-base": "^2.1.5",
    "react": "16.0.0-alpha.12",
    "react-native": "0.45.1"
  },
```

fbacker:
My Application has a depth structure of subpages using **react-native-router-flux**. Maybe that can cause the issue?

![out](https://user-images.githubusercontent.com/4478361/27288264-7c68b51c-5506-11e7-8971-34825d3d5dfb.gif)

shivrajkumar:
@fbacker For ActionSheet to work, it has to be called inside `Container` component of NativeBase

fbacker:
Sorry, I don't understand this.

I'm calling ActionSheet from a component. Now I've also tried to bubble up the event to a component that has Container -> Content in the render(). Still same issue.

shivrajkumar:
@fbacker Not able reproduce the issue on our end.
Can post the code snippet of it. Or create a test repo with the issue for us to look into it.

fbacker:
@shivrajkumar Will try to sit down in a couple of days and create an empty project and add router-flux-navigation and reproduce the issue. Thanks.

sankhadeeproy007:
@fbacker If your App is ios-only, you might as well use the native actionsheet that comes bundled with react-native.

fbacker:
Sorry no, both ios and android. It works for me to put the ActionSheet component in the render view. Maybe only me that has this issue :)

[...]

Issue #1121, opened 2017-07-29
Toast Issue with React-navigation close app


eggybot:
Hi Guys, I'm not sure why this is happening, this issue is link on this error 
#836 evaluating 'this.toastInstance._root' Error when using Toast 
https://facebook.github.io/react-native/docs/upgrading.html

First of all I'm using the latest version of NativeBase and ReactNative. The Toast is perfectly working but there is one issue I found which the error `evaluating 'this.toastInstance._root' Error when using Toast ` was showing when I do this steps.

First I newly open my app, then tried to test out the toast which perfectly works. Then I tried to go on different screen using the `react-navigation` and still it works fine. Then the error start I'm using this code.
```
        this.props.navigation.dispatch(NavigationActions.reset({
        index: 0,
        actions: [ NavigationActions.navigate({ routeName: 'Home'}) ]
        }));
```
As you can see, the back History `index:0,` is set to none for my navigation using react-navigation. In this setup if I use my Android back button, it will close the App. then if I open again my app, and tried to check out the Toast, the error shows up. It seems the `<Root>` tag on my main App is disappear which the error occur when I test the Toast again.

Do you know how can I fix this, every time I reopen the App, the error shows even I put the <Root> tag on my main component.

thanks,

sankhadeeproy007:
Are you just using the `StackNavigator`?

eggybot:
@sanketsahusoft 

yes, I'm using `stackNavigator` for `react-navigation`. It the same as the sample Toast in you documentation. But using  the `this.props.navigation.dispatch` and setting index to 0, the error shows once you totally use the back button on the phone which close the application then when you go back it seems the `Root` has been remove.

I cannot upload right now in github for a sample project which has thiss issue but you can [download ](http://resicert.com/oaptplf/toast_issue.zip) this small size code. Just create a project and just modify the app name so it will work.

eggybot:
any one can give me advice on this issue?

eggybot:
Hi Guys, just an update when I upraded my React Native it shows a new error which is
`Cannot read property 'showToast' of null`

can you check note me my this error show?

ohtangza:
I'm also experiencing `this.toastInstance._root'` error too...

eggybot:
This is the screenshot error
https://www.screencast.com/t/Gt6QCRcFs

[...]

PR #1700, opened 2018-03-16
If application root is reinitialized, update ActionSheet and Toast refs

fredmon3:
This PR is related to issues #1397, #1315, #937.

There is potential that React Native may unmount and remount the root node of the application due to Android lifecycle events without clearing the JS execution environment.  (This is possible even if native-base's `Root` is the top level component.)  This results in `actionSheetInstance` being invalid as it pointed to the previously mounted application root's ActionSheet component that was unmounted.  

This pull request updates the static `actionSheetInstance` and `toastInstance` refs regardless of what state they are in such that stale refs will be overwritten appropriately.

According to the [React Ref documentation](https://reactjs.org/docs/refs-and-the-dom.html):
> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts.

Thus, there is a null check on `c` in the ref callbacks.

### DIFF
--- a/src/basic/Root.js
+++ b/src/basic/Root.js
@@ -14,12 +14,12 @@ class Root extends Component {
         {this.props.children}
         <Toast
           ref={c => {
-            if (!Toast.toastInstance) Toast.toastInstance = c;
+            if (c) Toast.toastInstance = c;
           }}
         />
         <ActionSheet
           ref={c => {
-            if (!ActionSheet.actionsheetInstance)
+            if (c)
               ActionSheet.actionsheetInstance = c;
           }}
         />
--- a/dist/src/basic/Root.js
+++ b/dist/src/basic/Root.js
@@ -14,12 +14,12 @@ _react2.default.createElement(_reactNative.View,_extends({ref:function ref(c){re
 this.props.children,
 _react2.default.createElement(_ToastContainer.ToastContainer,{
 ref:function ref(c){
-if(!_ToastContainer.ToastContainer.toastInstance)_ToastContainer.ToastContainer.toastInstance=c;
+if(c)_ToastContainer.ToastContainer.toastInstance=c;
 },__source:{fileName:_jsxFileName,lineNumber:15}}),
 
 _react2.default.createElement(_Actionsheet.ActionSheetContainer,{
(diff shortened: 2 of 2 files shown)


nachofregeiro:
Will this pull request be approved soon to fix both ActionSheet and Toast issues?

fredmon3:
If there's any additional information that I can provide to assist with this pull request, please let me know. Thank you!

SunburtReynolds:
Is there an ETA on this PR?

adamawang:
+1

akemys:
+ 1

akemys:
+ 1

NappyPirate:
+1

akhil-geekyants:
Fixes #1397 #937 #1121 #1895

src/basic/Root.js
```diff
@@ -14,12 +14,12 @@ class Root extends Component {
         {this.props.children}
         <Toast
           ref={c => {
-            if (!Toast.toastInstance) Toast.toastInstance = c;
+            if (c) Toast.toastInstance = c;
```
douglasjunior: I think you don't need to check if `c` is different of null.

schie:
@akhil-geekyants @douglasjunior @fredmon3 any updates?

src/basic/Root.js
fredmon3: Thanks for taking a look at this @douglasjunior. Does GeekyAnts have a preference on how to handle the case where [`c` is `null`](https://reactjs.org/docs/refs-and-the-dom.html)? @akhil-geekyants @SupriyaKalghatgi

douglasjunior: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts. ref callbacks are invoked before componentDidMount or componentDidUpdate lifecycle hooks.

I think if the component has been unmounted, you should not keep the ref. This is memory leak.

SupriyaKalghatgi:
This week 💯

tapir:
Having the same issue as @gogulanareshkw.
Trying to show actionsheet twice after selecting an option triggers it

Issue #1706, opened 2018-03-20
Toast without timeout


danieloprado:
There is any way to set a toast without a timeout? Hide just after a user interaction?

akhil-geekyants:
@danieloprado no. Toast has a default timeout of 1500ms. It has been designed to hide after the timeout or button press whichever is earlier.
However, you can fork the library and change the code to meet your requirement.

PR #1737, opened 2018-03-27
Toast without timeout and reason for onClose (user or timeout)

danieloprado:
If duration is equal to 0 disable the toast timeout to close.
Added the reason that toast was closed: "user" when user press the button or "timeout" where duration ends.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1061,7 +1061,7 @@ declare module "native-base" {
 			position: "top" | "bottom" | "center";
 			type?: "danger" | "success" | "warning";
 			duration?: number;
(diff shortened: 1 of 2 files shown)


danieloprado:
Fix #1706

danieloprado:
Any updates?

NitroGhost:
any updates on this?

akhil-geekyants:
@danieloprado onClose is called twice with reason user and timeout, when duration is specified.

toast

NitroGhost:
@akhil-geekyants the whole Toast component needs rethinking really. there's a few issues surrounding this

PR #1830, opened 2018-04-24
performance improvement for tabs scrolling

slicejunk:
this makes the style compatible with scrollValue defined with ```"useNativeDriver": true```, 
```left``` throws an error.


SupriyaKalghatgi:
@slicejunk Please share GIF of this use-case, with existing NativeBase Tabs vs your fix

SupriyaKalghatgi:
@slicejunk Would you like this PR to be merged?

Issue #1847, opened 2018-04-28
Bar on side of itemlist component


kaym0:
All of the item lists have a bar on the side which indents them, but when making items into buttons (like in a drawer) it looks really stupid. Could you possibly add the option of taking this out? I understand that it works fine for the standard white design, but if you use any other design and want to highlight buttons like the current route it just looks super awkward having that bar there.

You can fix it using margins but it would be better if you could just not have an unneccessary indent there with another option. Like

```
<ListItem  noIntent> .... </ListItem>
```

Even the indent is fine if it's also styled by the listitem, because it would mean the awkward bar of none-listitem color wouldn't be there.
<img src="https://gyazo.com/a852a1445ef899835c17b3641ec1f034.png"/>

kaym0:
Fix for this is to set items as itemDivider such as below

```
<ListItem itemDividier style={{backgroundColor: "#000000"}}>
....
</ListItem>
```

You can setup background color for each item and it spans the entire length of the actual list 


<img src="https://gyazo.com/a9df3b99e094174e6859bf77607ef8b5.png"/>

SupriyaKalghatgi:
Fixed with [2.4.4](https://github.com/GeekyAnts/NativeBase/releases/tag/v2.4.4)

PR #1863, opened 2018-05-04
Added noIndent props to listItem

akhil-geekyants:
Enhancement based on #1847


SupriyaKalghatgi:
@akhil-geekyants No need of adding new prop `noIndent`
This should be with default List

PR #1894, opened 2018-05-14
Removed placeholderIconColor props of Picker

akhil-geekyants:


ch-tactica:
Howcome this was removed?

Issue #1825, opened 2018-04-23
Is it possible to optionally disable Picker button based on a condition?


VanessaChu:
How can I modify the Native Base Picker code to disable the button based on conditions?

SupriyaKalghatgi:
@VanessaChu You can try that yourself
NativeBase Picker is implementation of React Native Picker

ekimlinger:
This issue should not have been closed since this does not work on iOS.

@VanessaChu According to the RN docs, you can disable the input for their picker by passing an `enabled` flag, which works just fine on android. 

This however does not work on iOS and there is not currently a way to disable it.

SupriyaKalghatgi:
@ekimlinger Your statement sounds invalid

As said earlier
> NativeBase Picker is implementation of React Native Picker

![image](https://user-images.githubusercontent.com/13464885/40100888-3744ffca-5903-11e8-9842-f8a072dd7b99.png)

Since React Native supports this only for Android, we cannot help in adding this to iOS

SupriyaKalghatgi:
Released with [2.5.0](https://github.com/GeekyAnts/NativeBase/tree/v2.5.0)

PR #1899, opened 2018-05-16
Added prop enabled to picker for iOS

akhil-geekyants:
Enhancement based on issue #1825


SupriyaKalghatgi:
@akhil-geekyants Please share use-case here

akhil-geekyants:
enabled={false}

enabled false

enabled={true}

enabled true

SupriyaKalghatgi:
@akhil-geekyants Thank you

PR #1902, opened 2018-05-16
Adding icon typing to button component

darkmantle:
The prop typing on button is missing "Icon" which is causing annoying TS errors.


darkmantle:
Added noIndent typing for ListItem

SupriyaKalghatgi:
#1967

PR #1908, opened 2018-05-18
TS: Add all valid props to Content interface

emily-curry:
Based on the [react-native-keyboard-aware-scroll-view docs](https://github.com/APSL/react-native-keyboard-aware-scroll-view) (which the Content component wraps), this component should be able to accept all ScrollView props, as well some additional props to control the keyboard-aware-scroll-view (which have been directly copied from [that project's d.ts](https://github.com/APSL/react-native-keyboard-aware-scroll-view/blob/master/index.d.ts), as it does not export it's props). All redundant props have been removed from the interface.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -155,21 +155,25 @@ declare module "native-base" {
 		/**
          * see Widget Content.js
          */
-		interface Content {
-			/**
-             * The theme prop can be applied to any component of NativeBase.
-             */
-			refreshing?: boolean;
-			refreshControl?: object;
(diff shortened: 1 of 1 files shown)


SupriyaKalghatgi:
@Shurelia Can you explain me, why is `contentContainerStyle` and few other props excluded in your PR?

Shurelia:
Hey @SupriyaKalghatgi,
Those props are excluded because I changed `Content` to extend `ReactNative.ScrollViewProperties`. `contentContainerStyle` and those other excluded props are part of `ScrollViewProperties`.

PR #1944, opened 2018-05-28
Fix - Checkbox icon accept CheckboxIconMarginTop variable

patrick-welab:


SupriyaKalghatgi:
@patrick-welab Can you explain why is this needed?

patrick-welab:
@SupriyaKalghatgi I saw the theme variables `CheckboxIconMarginTop` is defined on https://github.com/GeekyAnts/NativeBase/blob/37558cb46ad11dcb97322d69aae1511cbc46e40b/src/theme/variables/platform.js#L90 but it is never used in the code
Also `CheckboxIconMarginTop` could be found on http://docs.nativebase.io/docs/ThemeVariables.html

SupriyaKalghatgi:
Thank you @patrick-welab for bringing this to our notice.
We should be looking into this with other perspective.

You will be updated on this here

PR #1970, opened 2018-06-07
Added expanded props to accordion.

akhil-geekyants:


SupriyaKalghatgi:
@akhil-geekyants Demonstrate this PR with example

akhil-geekyants:
@SupriyaKalghatgi 

expanded

jneljneljnel:
can this be programmatically changed? will the accordion re-render? for example, using dataArray.length -1   as the value for the expanded property does not auto expand the latest

jneljneljnel:
@SupriyaKalghatgi

PR #1987, opened 2018-06-11
Exposed onDateChange method for iOS

akhil-geekyants:


llezcano:
+1 to android support

akhil-geekyants:
On android there are only two actions 
1. dateSetAction - called when we click the ok button
2. dismissedAction - called when we click the cancel button

On iOS `onDateChange` is called every time the user change the date in the UI.

So adding `onDateChange` for android cannot be supported.

llezcano:
Thanks, I thought both (onDateChange and dateSetAction) were handling the same (a successfully closed datepicker). I was looking for that, maybe is a feature request.

ferreirix:
Sorry if I'm being really dumb but can't we at least call the `props.onDateChange ` when android fires `dateSetAction` ? Because the `dismissedAction` ins't supposed to change any state right?

PR #1996, opened 2018-06-12
Tab button text font size customizable

SupriyaKalghatgi:


YKSing:
Maybe should not override the `activeTextStyle` / `textStyle`?
I suggest put the `fontSize` before it instead of after it.

PR #1990, opened 2018-06-11
Fixed typo.

akhil-geekyants:
Fix for issue #1982  (Typo in onClose definition resulting in compilation error)

PR #1999, opened 2018-06-12
types: remove double semi-colon

NitroGhost:
the double semi-colon caused the following error: `node_modules/native-base/index.d.ts(1123,50): error TS1131: Property or signature expected`


SupriyaKalghatgi:
#1990

Issue #2005, opened 2018-06-13
Floating label onfocus issue


ViktorAksionov:
After upgrade to 2.5.1 from 2.4 floating label goes up only when user typed at least 1 char, when there is a focus in input field and nothing is there floating label stack in place, but should move up.

## Expected behaviour
![image](https://user-images.githubusercontent.com/2520890/41330037-e7710f82-6e85-11e8-8a47-7723ec2d95c2.png)

## Actual behaviour
![image](https://user-images.githubusercontent.com/2520890/41329976-945e41e8-6e85-11e8-8eba-f5e0bdc0dcd5.png)

## Is the bug present in both ios and android or in any one of them?
Checked only Android

ViktorAksionov:
![image](https://user-images.githubusercontent.com/2520890/41330756-27faacbc-6e8a-11e8-96c0-ed8d1ade4830.png)

ZeroCool00:
same problem here

akhil-geekyants:
@ViktorAksionov @ZeroCool00 can you try toggling the remote debugging mode.

SupriyaKalghatgi:
#2000

Issue #2000, opened 2018-06-12
floatingLabel glitch


theapache64:
Input with floatingLabel takes sometime to float the label.

### Code
```
import React, {Component} from 'react';
import {Container, Content, Form, Input, Item, Label} from "native-base";

export default class LabScreen extends Component {
    render() {
        return (
            <Container>
                <Content>
                    <Form>
                        <Item floatingLabel>
                            <Label>Sample label</Label>
                            <Input
                            />
                        </Item>
                    </Form>
                </Content>
            </Container>
        );
    }
}

```
### Output

![untitled](https://user-images.githubusercontent.com/9678279/41301419-8e92a5a6-6e85-11e8-943e-87b725376b7c.gif)

### I am on
```
"native-base": "^2.5.1",
    "react": "16.3.1",
    "react-native": "0.55.4",
```

SupriyaKalghatgi:
Share the device configuration
As in, model of device, which OS, which API etc

ZeroCool00:
same problem.. i have a nougat
![whatsapp image 2018-06-13 at 12 09 19 pm](https://user-images.githubusercontent.com/18139594/41334280-abcfe8ae-6f02-11e8-9728-3898619c584f.jpeg)

theapache64:
@SupriyaKalghatgi 

![screenshot_2018-06-13-12-58-59](https://user-images.githubusercontent.com/9678279/41336467-f964b44e-6f09-11e8-952f-117def3dc8af.png)

theapache64:
I've also tested it in my emulator which runs Android 7. Same issue. I think it's not device/os specific.

SupriyaKalghatgi:
we dd'nt face it ever
hence asked you for more details

ZeroCool00:
like i said.. it never happen before.. its just happen in this version **native-base": "^2.5.1**

akhil-geekyants:
@theapache64 @ZeroCool00 can you try toggling the remote debugging mode.

theapache64:
@akhil-geekyants Yeah, Turning off the debugging mode helped. (y)

akhil-geekyants:
Attaching a gif with `native-base` version 2.5.1

![floatinglabel](https://user-images.githubusercontent.com/27768184/41353665-8393a03c-6f3a-11e8-8b82-3a0cdf561055.gif)

theapache64:
It worked when i disabled the debug mode

ViktorAksionov:
[...]

PR #2030, opened 2018-06-19
Reverted fix 1606.

akhil-geekyants:
Fix for issue #2000 and #2005 based on issuecomment https://github.com/GeekyAnts/NativeBase/issues/2005#issuecomment-398262593


bhargav6744:
is this fix or still open ?

PR #2047, opened 2018-06-22
Added TimePicker

akhil-geekyants:


SupriyaKalghatgi:
timepicker

slavikdenis:
This seems promising! What's blocking it?

viniciusbo:
This would be really useful.

allan-simon:
Is it blocked because the CI is not passing ? if so it seems to not be because of the PR itself (the CI fails with "yarn test : test command not found"

shivrajkumar:
We are not adding any new components to the current version of NativeBase for now.

More details on #3069

Issue #1992, opened 2018-06-11
tabFontSize not working (material.js - ejectedTheme)


theapache64:

SupriyaKalghatgi:
Fixed with [v2.5.2](https://github.com/GeekyAnts/NativeBase/releases/tag/v2.5.2)

theapache64:
It's still not working. Am using v2.6.1. Here's the source code
[TabFontSizeTest.zip](https://github.com/GeekyAnts/NativeBase/files/2129792/TabFontSizeTest.zip)

theapache64:
Confirmed. It's working now (y)

PR #2051, opened 2018-06-25
Fix for issue #1992(tabFontSize not working)

akhil-geekyants:


SupriyaKalghatgi:
#1992

audiolion:
This change has unfortunately made it so `tabFontSize` of the theme always takes precedence over any `activeTextStyle` or `textStyle` object you explicitly pass to the `<Tab />` component. I feel like the order should be reversed so if you explicitly pass `fontSize` it should override the theme default.

PR #2062, opened 2018-06-27
feat: added toast onButtonClick functionality (requires documentation update)

barrymichaeldoyle:
I'm not sure how to updated the documentation. But basically this new feature allows the toast config to take an additional config key called `onButtonClick` which accepts any function that runs when the button is clicked, but not when the toast closes (it does however also close the toast upon clicking it after running the function).


barrymichaeldoyle:
Not sure why the build is failing :/

PR #2066, opened 2018-06-28
[Typescript] Add transparent prop to Header Interface

coman3:
Please refer to docs https://docs.nativebase.io/Components.html#header-transparent-headref


coman3:
was in development branch.... (thanks github search 😐 )

PR #2100, opened 2018-07-04
Removed Drawer from NativeBase

akhil-geekyants:


adaltavo:
Why this was removed?

SupriyaKalghatgi:
@adaltavo react-navigation is part of react-native, default navigation

stevefan1999:
I don't think it should be removed, people like me like to use `react-router-native` and the removal of `Drawer` is definitely unfavored. Thankfully we still have [root-two/react-native-drawer](https://github.com/root-two/react-native-drawer) as an alternative, it should be well compatible.

cylim:
I think it would be better to keep the drawer. The `react-navigation` drawer navigator is not as flexible as `react-native-drawer`.

spotsadmin:
The release note says Drawyer is deprecated - this is not a deprecation, but a complete removal with no warning. Changes like these cause undue burden on native-base users :(

flyandi:
Yeah this kind of sucks - you can't just remove a major component from the library without forewarning. Lots of people depend on this stuff or at least you need to give plenty of forewarning that you intend to so because of XYZ reason - that is fine - but not like this. Bad quality!

fuj1w4r4:
I am came to this at first I thought I am the only one facing this trouble. How can it be removed as it is part of an important UI features?

flyandi:
So what are you guys going to do about this? Just ignore this thread? 

For now I will be stop using NativeBase for my apps since this seems to be an amateur project how it is handled. Not only the documentation is updates which still references these components. New users will be highly confused not to speak the plenty of apps that are relying on this feature.

Sorry guys but this is not really acceptable if you position yourself as an essential dependency.

As a remedy I created the `native-base-legacy` repo that collects these things here so you will still be able to compile your current app.

Head over to http://github.com/flyandi/native-base-legacy or just ```yarn add native-base-legacy```

then

```
import { Drawer } from 'native-base-legacy'
```

and you should be good to go.

sanketsahusoft:
Everyone, we are very sorry about this. We are going to add it back!

There were issues with the Drawer component coming from a third party library. We are going to revert this change and fix the issues of react-native-drawer or provide an alternative solution.

SupriyaKalghatgi:
Added Drawer back with [2.8.0](https://github.com/GeekyAnts/NativeBase/releases/tag/v2.8.0)

flyandi:
Thank you!

Zyphxion:
Thank you!

PR #2110, opened 2018-07-06
Upgrading react-native-vector-icons dependency

bernerbrau:
Attempting to use Octicons in NativeBase currently either doesn't work, or displays the wrong glyph, due to the glyph map in the dependency and the font bundled in NativeBase being out of sync.

This PR updates the Octicons font to the latest one provided by react-native-vector-icons, and it also uses the latest RNVO dependency.


SupriyaKalghatgi:
@bernerbrau did you check the latest release and it's release notes

SupriyaKalghatgi:
@bernerbrau Pushed this on development
https://github.com/GeekyAnts/NativeBase/commit/4420ec37ee8e802b38237703e82824307fcf7d13
https://github.com/GeekyAnts/NativeBase/commit/9963c3e2055fcbefb58a2d9df87e655fb43c1409

Issue #2113, opened 2018-07-09
Date Picker : set as disabled


blankonn:
Versions : 

- NativeBase : 2.6.1

Issue : 
I need to set the DatePicker to disabled so the user will be unable to make a selection. However, I couldn't find the props `enabled` in DatePicker which allow me to set the DatePicker to be disabled. is this will be implemented or yet to be implemented ?

Thanks in Advance

PR #2114, opened 2018-07-09
Added enabled props to datepicker

akhil-geekyants:
PR for Issue #2113 (Date Picker : set as disabled)


SupriyaKalghatgi:
@akhil-geekyants NativeBase so far has `disabled` props to enable / disable the components
Lets follow same standards

PR #2171, opened 2018-07-26
FIXES Item's onBlur event passing;

fatalVoltage:
Actually passes an event object through the onBlur method in `Item`, since some libraries like [Formik](https://github.com/jaredpalmer/formik) need to pass said event to function properly.


SupriyaKalghatgi:
@fatalVoltage Need use-case to show how NativeBase Item `onBlur` functions with and without this PR

PR #2212, opened 2018-08-16
Picker modal anim

kousun12:
add a modal animation style prop for the picker's modal


suvenduchhatoi:
@kousun12 There is already slide animation in Picker, as per design intention

PR #2243, opened 2018-08-28
Revert "Removed Drawer from NativeBase"

SupriyaKalghatgi:
Reverts GeekyAnts/NativeBase#2100


trickpasp:
I'm having trouble with drawer, I can not use scrowview with it, could you help me?

PR #2254, opened 2018-09-01
Fixes #2213

### LINKED ISSUE #2213: Swipe List Error when changing data source
Hi!, i having a problem with swipeable List i couldn't figure out how to resolve
Im using a code like this

`
list = () => {
		return (
			<List
				bounces={false}
				dataSource={this.ds.cloneWithRows(
					this.props.trackerData[this.state.category]
				)}
				renderRow={this.renderRow}
				disableRightSwipe
				renderRightHiddenRow={(data, secId, rowId, rowMap) => {...}}
				rightOpenValue={-130}
			/>
		)
	}
`

Here is an image of my View
Im using list view to render elements from an array of data, an lets say, i render 2 elements in this list, AND swipe LAST element an let it open.

whatsapp image 2018-08-16 at 4 03 59 pm

then if i change category in my state, it means it change the data source of the list, and the new list has fewer elements than the previous one, AND SWIPE an element lets say the FIRST ONE, im getting this error.

whatsapp image 2018-08-16 at 4 03 59 pm 1

it happens when the index of the previous opened element, is higher than the length of the actual datalist. 
 im getting
`null is not an object (evaluating 'this._rows[this.openCellld]._root')`

i have already tried lot of things like closing the row with rowMap or obtaining reference to inner elements like openCellId, but cant find solution to this.

it does not happen when i delete an element from the list and close it with 		`rowMap[`${secId}${rowId}`].props.closeRow()`

sdacunha:
Fixes issue #2213 on SwipeableList/SwipeRow where a deleted row would cause a crash due to safeCloseOpenRow() being called on a a row @ openCellId that no longer exists.


suvenduchhatoi:
@sdacunha  Found no error while reproducing #2213 
screen shot 1440-06-08 at 12 27 59 pm
screen shot 1440-06-08 at 12 28 42 pm

PR #2260, opened 2018-09-04
add small padding to ListItem

Mahdeep:


SupriyaKalghatgi:
Can you share the use case for this? In what case is this needful?

PR #2283, opened 2018-09-19
enhancement/add support for iPhoneXR

Piotrovskyi:
Add support for iPhoneXR
iPhone XS look good even with old code

### DIFF
--- a/src/theme/variables/commonColor.js
+++ b/src/theme/variables/commonColor.js
@@ -7,7 +7,7 @@ const deviceWidth = Dimensions.get("window").width;
 const platform = Platform.OS;
 const platformStyle = undefined;
 const isIphoneX =
-  platform === "ios" && (deviceHeight === 812 || deviceWidth === 812);
+  platform === "ios" && (deviceHeight === 812 || deviceWidth === 812 || deviceHeight === 896 || deviceWidth === 896);
 
 export default {
   platformStyle,
--- a/src/theme/variables/material.js
+++ b/src/theme/variables/material.js
@@ -7,7 +7,7 @@ const deviceWidth = Dimensions.get("window").width;
 const platform = Platform.OS;
 const platformStyle = "material";
 const isIphoneX =
-  platform === "ios" && (deviceHeight === 812 || deviceWidth === 812);
+  platform === "ios" && (deviceHeight === 812 || deviceWidth === 812 || deviceHeight === 896 || deviceWidth === 896);
 
 export default {
   platformStyle,
--- a/src/theme/variables/platform.js
+++ b/src/theme/variables/platform.js
@@ -7,7 +7,7 @@ const deviceWidth = Dimensions.get("window").width;
 const platform = Platform.OS;
 const platformStyle = undefined;
 const isIphoneX =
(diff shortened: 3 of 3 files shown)


SupriyaKalghatgi:
This doesn't fix the issue

jorgebaralt:
Please make sure this supports also iphone XS Max

craftzdog:
confirmed it works.

alexisbronchart:
Changing the `isIphoneX` condition in the themes is not enough if you have an ejected theme like I have. There's actually a bug in `Footer.js` and `Header.js`. See lines https://github.com/GeekyAnts/NativeBase/blob/b7d1ebce89c4533e71a03590de2eb9e2bfdbd646/src/basic/Footer.js#L84 & https://github.com/GeekyAnts/NativeBase/blob/b7d1ebce89c4533e71a03590de2eb9e2bfdbd646/src/basic/Header.js#L83 the check is wrongly made on `variable` instead of `variables` which takes the context theme into account.

jorgebaralt:
none of the changes in this discussion change the way header is behaving

rreina:
I've tested all proposed solutions. None of them are working. 

This was our solution:
Considering iPhone X, XS, XS Max and XR have the same screen aspect ratio in common. We have added our custom code to detect that:
`    
   private readonly iPhoneXAspectRatio = 2.16;
   ...
    hasiPhoneXAspectRatio = (): boolean => {
        if (this.runningOniOS) {
            const screenAspectRatio =  Number((this.ScreenHeight / this.ScreenWidth).toFixed(2));
            return screenAspectRatio === this.iPhoneXAspectRatio;
        }
        return false;
    }
`
Then, we have ejected the NativeBase theme and removed iPhone X support (in our case in Platforms.js) by setting all iPhone X safe area Insets = 0. This was the only way we found to invalidate iPhone X support in NativeBase because the 'Platform.isIphoneX' ejected variable is not working at all.

Finally, we have used our custom iPhoneX like detection to add a safe area view in our App main view for these devices.

jorgebaralt:
@Piotrovskyi solution worked for me.

alexisbronchart:
@rreina The isIphoneX variable is not working because of the bug I mentioned in my previous message. There is another PR that will fix it #2383

rreina:
@alexisbronchart I agree your proposed fix should be integrated to allow overriding the 'isIphoneX' variable. Either way, you can always set all iPhone X safe area Insets = 0 in the ejected theme and implement your own iPhone XS, XS Max and XR detection and UI adaptations. 

For us, it was not enough adapting the header and the footer of the application. Some full screen components (modal components above all) which are not wrapped by FooterTabs, needed a safe zone as well. This was the mayor reason to not use the adaptations provided by NativeBase.

gameboyVito:
when to merge?

jorgebaralt:
why haven't this been merged?

craftzdog:
Seems like it's been resolved by below commit and already shipped as `^2.9.2`:

https://github.com/GeekyAnts/NativeBase/commit/9e324204be9699c51d14957ec55835c7a2f97487

PR #2318, opened 2018-09-30
fixed transparent in CardItem

Mahdeep:


suvenduchhatoi:
The design specification is intended in a way to to keep the background transparent for cardItems rendered inside of a transparent card. So, there is no need of styling cardItems rendered inside a transparent card.

PR #2332, opened 2018-10-04
Add picker property to Item

### LINKED ISSUE #2325: Missing Typescript definition for picker property of item

### node, npm, react-native, react and native-base version, expo version if used
node v8.11.3
npm 6.1.0
expo sdk 29
react-native 0.55.4
react 16.3.1
native-base 2.8.0

### Expected behaviour
Should be able to set picker property of item when child of Form component

### Actual behaviour
typescript error due to missing definition

### Steps to reproduce
Add basic form and item and attempt to set picker property

wddelacruz:
This should fix #2325

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -796,6 +796,7 @@ declare module "native-base" {
 			success?: boolean;
 			last?: boolean;
 			style?: ReactNative.ViewStyle | Array<ReactNative.ViewStyle>;
+			picker?: boolean;
 		}
 
 		interface Form {


wddelacruz:
@burtek i ended up creating a script that just overwrote the original typings every time. I'ts a very dirty solution but it's been over a month now and i'm not sure if they actually go through the issues and pull requests of their repos.

burtek:
@wddelacruz As sad as it is, I'm probably gonna stick with `<Item {...{picker: true}} />` notation for now. Let's hope they'll get the typings updated. IMHO it should be a must to update typings when doing changes to JS codebase

SupriyaKalghatgi:
#2473

Issue #2248, opened 2018-08-30
floatingLabel color not working?


baba43:
Hi,

I simply want to change the text color of my inputs. The source looks like this:

```
<Item floatingLabel>
  <Label style={{ color: '#ffffff' }}>Username</Label>
  <Input style={{ color: '#ffffff' }} />
</Item>
```

However, on my Android device the color change is only applied as long as the placeholder is not floating. As soon as I start typing, the placeholder starts floating and the color is resetted.

Android: 8.0.0
native-base: 2.8.0

SupriyaKalghatgi:
Please follow issue guidelines
Failing which this issue will be closed

Moreover, this query is answered on Stack Overflow

SupriyaKalghatgi:
https://github.com/GeekyAnts/NativeBase/issues/1964
https://stackoverflow.com/questions/48123684/change-react-native-base-floating-label-label-color

baba43:
Thank you for the answer. However, I have already found that thread and as you can see, my code already looks like that. The problem is, that it is not working.

SupriyaKalghatgi:
@baba43 If you check the above links, it has output sample demonstrating floatinglabel color

But then due to https://github.com/GeekyAnts/NativeBase/pull/2001, this caused issue

PR #2387, opened 2018-10-23
Fix floatingLabel animation

elyalvarado:
This is commit fixes the expected animation for the floating label,
which should change the fontSize and line height when the field is
focused.

This is basically reverting commit [3e3ba4e](https://github.com/GeekyAnts/NativeBase/commit/3e3ba4e1a273da9b3236e0f6696e3873152d4a8f#diff-f1939c40e767531823521e714e4db9c0), which was theoretically fixing #2248, but the true is that the root cause of that issue was not in that line, and instead it caused a regression were the floatingLabel is not resizing to an smaller size on focusing the input.

The current behaviour without this fix is:

current

The expected behaviour is (making the label smaller when focusing):

expected

With this fix you're still able to change the color of the label by applying style directly to it (which I understand was what initially was reported as an issue in #2248):

expectedcolor


elyalvarado:
@SupriyaKalghatgi can you take a look at this? Is a trivial fix and it affects the UX for people that upgraded to v2.8.1

elyalvarado:
@Angelk90 there are two ways that should work, but the behavior will depend on the version of native-base you are using:
1. Add the color style directly to your `Label` in this way: `<Label style={{color: 'blue'}}>`.
  1.1 If you're using 2.8.1 this will change the label color for both floating/not floating, but the font-size won't decrease when floating.
  1.2 If you're using 2.8.0 this will change the label color for not floating, but not when floating
  1.3 If this commit is merged it'll behave as expected
2. Eject the theme and change the `inputColorPlaceholder` property.
  2.1 If you're using 2.8.1 it'll behave the same as point 1.1 above
  2.2 If you're using 2.8.0 it'll behave as expected
  2.3 If this commit is merged it'll behave as expected

elyalvarado:
@SupriyaKalghatgi I amended the commit, and it now fixes #2248 and removes the regression for not changing the label font size when floating. The root cause of the issue was that StyleSheet.flatten should have been called with an array, otherwise it just returns the first argument and therefore was disregarding the actual Label styles coming from the props.

Angelk90:
@elyalvarado : The button in the input field for the clear? Can you post an example, for example on snack.expo.io?

elyalvarado:
@Angelk90, I see what you mean now. 

The button clearing the fields that you see at the right end of it in my gifs is the native iOS clear button, which I enabled by passing ` clearButtonMode: 'while-editing'`  as prop to the `Input`, which in turns passes it to the underlying react-native `TextInput` component (see [React Native Documentation](https://facebook.github.io/react-native/docs/textinput#clearbuttonmode).

Unfortunately, that property is iOS specific, and you can't style it.

SupriyaKalghatgi:
@elyalvarado I will take a look on this

iamsokol:
@elyalvarado @SupriyaKalghatgi
Why doesn`t label  font size decrease on input focus? What can I do with it?
"native-base": "2.10.0",
"react-native": "0.57.7",

exemple

elyalvarado:
@iamsokol even though the PR was already merged, it was merged before the release of the 2.10.0 version of `native-base`. It'll be sure included in the next version release.

iamsokol:
@elyalvarado As I understand it is already released (as I saw versions fix),but it doesn't work in my code.

elyalvarado:
Is merged, but not yet released, take a look at the list of [commits](https://github.com/GeekyAnts/NativeBase/commits/master) and you'll see that the merge commit is just after the one tagged 2.10.0

iamsokol:
2.8.2
> Item: FloatingLabel animation issue fixed.

@elyalvarado What is this fix about!Can it help me?

elyalvarado:
@iamsokol that fixed another issue with the floating label color, but introduced the size issue.

iamsokol:
@elyalvarado can I change label color, when input has focus?

PR #2423, opened 2018-11-06
Fix Footer when height equals 0

max-konin:
Hi. I faced with the issue  that when I create a footer with height == 0 like:
```javascript
<Footer style={{ height: 0 }} />
```
The footer is rendered with default height. 

Footers with zero height can be useful for example when I adapt a screen w/o footer for iPhoneX.


SupriyaKalghatgi:
Cannot accept test cases

max-konin:
@SupriyaKalghatgi why?

max-konin:
@SupriyaKalghatgi Could you describe a reason to close PR?

PR #2445, opened 2018-11-16
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
Test against all current LTS and stable branches of Node.js.
Testing single versions of Node.js makes not much sense as there are often updates.


tux-tn:
There is no CI pipeline

DanielRuf:
Well, I just updated the CI config.

PR #2479, opened 2018-12-04
Tabs rtl fix

idealamz:
when using 

```
import { I18nManager } from 'react-native'
I18nManager.forceRTL(true)
```
tabs header and content are messy.

this PR should fix it.


ansarikhurshid786:
I am using latest version of nativebase header and content working but header indicator is messing.

ansarikhurshid786:
tab underline is wrong.

Issue #2490, opened 2018-12-16
Accordion property `expanded` does not exist in Typescript


burtek:
## I have gone through these following points
- [x]   Check latest documentation: https://docs.nativebase.io/
- [x]   Check for existing open/closed issues for a possible duplicate before creating a new issue: https://github.com/GeekyAnts/NativeBase/issues
- [x]   Use the latest NativeBase release: https://github.com/GeekyAnts/NativeBase/releases
- [x]   Check examples from NativeBase KitchenSink https://github.com/GeekyAnts/NativeBase-KitchenSink
- [ ]   For discussion purpose make use of NativeBase Slack: http://slack.nativebase.io/
- [ ]   For queries related to theme, check [Theme Variables](https://docs.nativebase.io/docs/ThemeVariables.html) from Docs and live NativeBase Theme Editor http://nativebase.io/customizer/

## Issue Description
[Accordion](https://docs.nativebase.io/Components.html#accordion-def-headref)'s typescript typings don't have `expanded` property. Using `native-base: 2.8.1`.

https://github.com/GeekyAnts/NativeBase/blob/e96c997492d898d812243d7f19297189c75af4c0/index.d.ts#L845-L856

burtek:
I've prepared PR #2491 which among others fixes this issue

burtek:
@SupriyaKalghatgi any chance for getting that PR merged? 🤔

PR #2491, opened 2018-12-16
[TS] Add `expanded` prop to Accordion and allow Accordion's item strong-typing

burtek:
This PR fixes https://github.com/GeekyAnts/NativeBase/issues/2490.

By the way I also used generics to strong-type Accordion item type.


SupriyaKalghatgi:
Cannot have strong-type
Fixed with commit https://github.com/GeekyAnts/NativeBase/commit/7b8a09b2bc9d70ef8f0e20a2daa30002ea45a91d

burtek:
Right. Mind explaining why?

PR #2510, opened 2019-01-03
Make DatePicker use disabled prop (fix #2509)

### LINKED ISSUE #2509: DatePicker disabled prop does not update component state
- [x]   For discussion purpose make use of NativeBase Slack: http://slack.nativebase.io/
- [x]   For queries related to theme, check [Theme Variables](https://docs.nativebase.io/docs/ThemeVariables.html) from Docs and live NativeBase Theme Editor http://nativebase.io/customizer/

## Issue Description
The `disabled` prop on `DatePicker` only works on component mounting, this is the property is set initially but when later changed the component remains in the previous state

### node, npm, react-native, react and native-base version, expo version if used, xcode version

node: v8.12.0
yarn: v1.12.3
react: 16.6.1
react-native: 0.57.7
native-base: v2.10.0
Xcode: v10.1

### Steps to reproduce
Run the following snippet and check that when disabled is changed to `false` the DatePicker remains disabled.

```
import React from 'react'
import { Container, Header, Content, Text, Button, DatePicker } from 'native-base'

export default class DatePickerTest extends React.Component {
  state = { disabled: true }

  toggleState = () => {
    const disabled = !this.state.disabled
    this.setState({ disabled })
  }

  render() {
    const { disabled } = this.state

    return (
      <Container>
        <Header/>
        <Content>

          <Text>
            Current State:
            <Text style={{backgroundColor: disabled ? 'red' : 'green'}}>
              { disabled ? 'disabled' : 'enabled' }
            </Text>
          </Text>

          <DatePicker
            defaultDate={new Date(2018, 4, 4)}
            minimumDate={new Date(2018, 1, 1)}
            maximumDate={new Date(2018, 12, 31)}
            locale={"en"}
            timeZoneOffsetInMinutes={undefined}
            modalTransparent={false}
            animationType={"fade"}
            androidMode={"default"}
            placeHolderText="Select date"
            textStyle={{ color: "green" }}
            placeHolderTextStyle={{ color: "#d3d3d3" }}
            disabled={disabled}
          />

          <Button block light onPress={this.toggleState} >
            <Text>Toggle enabled/disabled state</Text>
          </Button>

        </Content>
      </Container>
    )
  }
}
```

### Expected behavior
Changing the `disabled` prop should change the component state

The issue is present in both platforms

elyalvarado:
This commit fixes issue #2509.

Previous to this change the disabled prop on DatePicker only worked on
component mounting, this is the property was set initially but when
later changed the component remained in the previous state.

This commit also removes two calls to `setState` on `componentDidMount`
which prevents an innecesary call to `render` on mounting.


suvenduchhatoi:
@elyalvarado As per PR protocol, please do this
screen shot 1440-06-02 at 11 40 00 am

PR #2514, opened 2019-01-06
Allow iPhoneX feature flag to be modified from specified theme (instead of default)

### LINKED ISSUE #2382: Users cannot change `isIphoneX` variable in custom themes.
- [x]   For discussion purpose make use of NativeBase Slack: http://slack.nativebase.io/
- [x]   For queries related to theme, check [Theme Variables](https://docs.nativebase.io/docs/ThemeVariables.html) from Docs and live NativeBase Theme Editor http://nativebase.io/customizer/

## Issue Description

I am no able to modify the `isIphoneX` variable in my custom theme.  Since Native Base currently does not support iPhone XS, iPhone XR and iPhone XS Max, may app does not render properly on those phones.

I think this was an oversight and can be corrected by adding an `s` to three files.  People with custom themes should be able to modify this variable.

### node, npm, react-native, react and native-base version, expo version if used
node v8.11.1
npm v6.4.1
react-native 0.56
native-base 2.8.0
expo not used

### Expected behaviour

Add the following to your local project's `native-base-theme/variables/platform.js`:

```
import DeviceInfo from 'react-native-device-info';
const isIphoneX = DeviceInfo.hasNotch();  // This can be any custom behavior and does not need to rely on DeviceInfo
```

We would then expect the `Header`, `Footer` and `Content` components to respond appropriately to the value of the `isIphoneX` variable set above.  

### Actual behaviour

The `Header', 'Footer' and 'Content' components do not response to the overridden value of `isIPhoneX`.

### Steps to reproduce

Install `react-native-device-info`

```
npm install --save react-native-device-info
```

Add the following to your local project's `native-base-theme/variables/platform.js`:

```
import DeviceInfo from 'react-native-device-info';
const isIphoneX = DeviceInfo.hasNotch();
```

Run your app in an iPhone XR simulator.  You will notice that the `Header`, `Footer` and `Content` components do not provide adequate margins for the newer iPhones.

It is present in both.

### Any other additional info which would help us debug the issue quicker.
The issue is easy to fix.  Simple change `variable.isIphoneX` to `variables.isIphoneX` in the `Header.js`, `Content.js` and `Footer.js` files.

I will open a PR.

confiks:
Fixes #2382. Similar to #2383, which was incorrectly closed. Now without committing dist files.

### DIFF
--- a/src/basic/Content.js
+++ b/src/basic/Content.js
@@ -91,7 +91,7 @@ class Content extends Component {
     const variables = this.context.theme
       ? this.context.theme["@@shoutem.theme/themeStyle"].variables
       : variable;
-    return variable.isIphoneX ? (
+    return variables.isIphoneX ? (
       <KeyboardAwareScrollView
         automaticallyAdjustContentInsets={false}
         resetScrollToCoords={
--- a/src/basic/Footer.js
+++ b/src/basic/Footer.js
@@ -81,7 +81,7 @@ class Footer extends Component {
     const variables = this.context.theme
       ? this.context.theme["@@shoutem.theme/themeStyle"].variables
       : variable;
-    return variable.isIphoneX ? (
+    return variables.isIphoneX ? (
       <View
         ref={c => (this._root = c)}
         {...this.props}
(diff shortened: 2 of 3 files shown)


suvenduchhatoi:
@confiks As per PR protocols, you need to include this
screen shot 1440-06-02 at 6 53 53 pm

PR #2553, opened 2019-02-04
Actionsheet fix

suvenduchhatoi:


shubham1164:
Does this fix the extra space at the bottom of the Action sheet? - Excess bottom space in the ActionSheet #2624

Issue #2176, opened 2018-07-29
Header error on native base web


codejamninja:
I am getting the following error when using the Header.js component on the web.

The header displays and operates correctly. It just throws the following error in the browser console.

```
index.js:38 Warning: Received `false` for a non-boolean attribute `uppercase`.

If you want to write it to the DOM, pass a string instead: uppercase="false" or uppercase={value.toString()}.

If you used to conditionally omit it with uppercase={condition && value}, pass uppercase={condition ? value : undefined} instead.
    in div (created by Text)
    in Text (at Text.js:30)
    in Text (at connectStyle.js:384)
    in Styled(Text) (at MainContent.js:26)
    in a (created by Link)
    in Link (at MainContent.js:25)
    in div (created by View)
    in View (at Header.js:102)
    in div (created by View)
    in View (at Header.js:67)
    in Header (at connectStyle.js:384)
    in Styled(Header) (at MainContent.js:19)
    in div (created by View)
    in View (at Container.js:15)
    in Container (at connectStyle.js:384)
    in Styled(Container) (at MainContent.js:18)
    in MainContent (at Home.js:47)
    in Home (created by Connect(Home))
    in Connect(Home) (created by Route)
    in Route (at index.js:12)
    in Switch (at index.js:10)
    in Routes (at App.js:6)
    in App (at Reaction.js:45)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Connect(ConnectedRouter))
    in Connect(ConnectedRouter) (at Reaction.js:44)
    in PersistGate (at Reaction.js:43)
    in Provider (at Reaction.js:42)
    in Reaction (at ClientApp.js:13)
    in StyleProvider (at ClientApp.js:12)
    in Router (created by BrowserRouter)
    in BrowserRouter (at ClientApp.js:11)
    in ClientApp
    in div (created by View)
    in View (created by AppContainer)
    in div (created by View)
    in View (created by AppContainer)
    in AppContainer
```

Here is how I am using the Header
```js
        <Header
          style={{
            alignItems: 'center',
            justifyContent: 'center'
          }}
        >
          <Link to="/home">
            <Text style={{ fontSize: 24 }}>{config.title}</Text>
          </Link>
        </Header>
```

Also, here is the theme styling I'm using for Header

```js
import { PixelRatio, StatusBar } from 'react-native';
import variable from '~/theme/variables/platform';

export default (variables = variable) => {
  const { platform, platformStyle } = variables;
[...]

PR #2619, opened 2019-03-02
Fix uppercase prop warning on web

oxyii:
#2176

### DIFF
--- a/src/basic/Text.js
+++ b/src/basic/Text.js
@@ -11,6 +11,7 @@ class Text extends Component {
     const {
       uppercase,
       children,
+      ...others
     } = this.props;
 
     let text;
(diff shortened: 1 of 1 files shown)


jmatsushita:
Also ran into this. @SupriyaKalghatgi Please merge.

SynerG:
This prevents the `uppercase` warning at `Text`, but is not the only one.
`note` is also drilled down to `Text`.
`active` is passed to `IconNB`.
All of them generate similar warnings as they are unexpected props at the final components.

oxyii:
@SynerG I'm ready to add your additions, and I know more places with similar issues. But I also see the lack of response from NB authors. So I don't want to waste my time on it if it is not claimed. I'm sorry. Closed

PR #2622, opened 2019-03-03
TS: Add KeyboardAwareScrollViewProps to Content interface

tamurayk:
TypeScript detects an error when add onScroll property to Content Component.

It seems Contents should extend `KeyboardAwareScrollViewProps`.

Similar to #1908,
actually It should add KeyboardAwareScrollViewProps instead of ScrollViewProperties.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,7 @@
 declare module "native-base" {
 	import * as React from "react";
 	import * as ReactNative from "react-native";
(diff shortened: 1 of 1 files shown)


tamurayk:
see: https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/Content.js#L95

ecbrodie:
Requesting attention from the Native Base team. It has been open without activity for a month now. Can we please get this simple change merged? @SupriyaKalghatgi

PR #2651, opened 2019-03-30
Fix for react-native-vector icons missing or wrongly mapped

ajaybhatia:


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "native-base",
+  "name": "@ajaybhatia/native-base",
```
sankhadeeproy007: I think you meant to add this only for your fork and not here.

package.json
```diff
@@ -66,7 +66,7 @@
     "react-test-renderer": "16.4.1"
   },
   "gitHead": "5bbeacc403ba97622703699132c55d8359344004",
-  "homepage": "https://github.com/GeekyAnts/NativeBase#readme",
+  "homepage": "https://github.com/ajaybhatia/NativeBase#readme",
```
sankhadeeproy007: I think you meant to add this only for your fork and not here.

package.json
```diff
@@ -77,7 +77,7 @@
   "readmeFilename": "README.md",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/GeekyAnts/NativeBase.git"
+    "url": "git+https://github.com/ajaybhatia/NativeBase.git"
```
sankhadeeproy007: I think you meant to add this only for your fork and not here.

shivrajkumar:
This is no longer relevant as the package is already updated and mapped.

PR #2676, opened 2019-04-23
color prop defined in Radio Interface

naxir:
After exploring how to change the outer circle color of radio button in native base i have found that in Radio.js file 
this.props.color is defined but not defined in Interface of Radio Button in the file i.e index.d.ts

So it should be defined so that intelligence should work


naxir:
@SupriyaKalghatgi  have a look on that PR 

Thanks :)

PR #2678, opened 2019-04-23
Add 'onDismiss' to ToastContainer

ahartzog:
Separate from 'onClose' which fires on all close events, onDismiss only executes on the button press of the toastContainer

### DIFF
--- a/src/basic/ToastContainer.js
+++ b/src/basic/ToastContainer.js
@@ -66,16 +66,20 @@ class ToastContainer extends Component {
       buttonTextStyle: config.buttonTextStyle,
       buttonStyle: config.buttonStyle,
       textStyle: config.textStyle,
-      onClose: config.onClose
+      onClose: config.onClose,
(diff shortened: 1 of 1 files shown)


ahartzog:
Per https://github.com/GeekyAnts/NativeBase/pull/1542 and some other issues I've seen (and I need this functionality in our project), just adds a separate callback function when the toast is dismissed by the user pressing the button.

Prettier also saved a couple other changes that appear to be consistent with the project formatting so I left them in place.

This is my first time contributing to NB so let me know if there's any additional standards or conventions to follow.

ahartzog:
Ah, actually, now that I'm in this I see that the "reason" is being passed back into the function passed in so I can use that to execute my logic accordingly.

ahartzog:
I'm going to close this out since what I needed is already built in, I'll see if there's a way to document this functionality easily so it can more readily apparent.

PR #2722, opened 2019-06-08
add onfocus event to textInput

mahyarfrd70:


sankhadeeproy007:
Thanks for the PR! But we are already passing forward all the props that are being sent. So that makes this redundant.

PR #2740, opened 2019-07-03
Button cleanup

Srishti-Sharma:


sanketsahusoft:
@sankhadeeproy007 We have the theme file. Instead of using constants, can we pull the defaults from the theme file instead?

Srishti-Sharma:
@sanketsahusoft updated.

PR #2756, opened 2019-07-09
Add missing Textarea type definitions

alexander7161:
Bordered and underline were missing from the type definition of Textarea.


Aryk:
@sankhadeeproy007 , @alexander7161 

Why are `rowSpan` `bordered` and `underline` required to be entered....?

It's causing me Typescript errors but works perfectly fine without me entering these values...

If it's an error, can you guys please add a `?` to the type definitions to make them optional?

PR #2760, opened 2019-07-10
Adding TabBar disabled feature

Srishti-Sharma:
Simulator Screen Shot - iPhone X - 2019-07-10 at 12 02 00

Tab3 is disabled.


src/basic/Tabs/DefaultTabBar.js
```diff
@@ -74,18 +84,20 @@ const DefaultTabBar = createReactClass({
...
           </TabHeading>
         </Button>
       );
-    } 
+    }
     return (
       <Button
         style={{ flex: 1 }}
+        disabled={isDisabled}
         key={_.random(1.2, 5.2)}
```
sankhadeeproy007: Replace random with the tab name here.

Srishti-Sharma: Okay

PR #2781, opened 2019-07-18
Add POSITION enum to ToastContainer

PlabanJr:


LuckyShine001:
Toast with position top is disturbing, when keyboard is opened.  top is getting value of keyboard height. 
only on iOS

PR #2793, opened 2019-07-22
Remove FontAwesome5Pro

PlabanJr:


roman-yerin:
Why did you remove FA5 Pro?

PR #2794, opened 2019-07-22
Fixes crashing of prop floatingLabel in Item

Srishti-Sharma:


stackerine:
The PR fix the problem thank.
However, in my opinion you should not disable native animation acceleration. We should replace `top` displacement by `transform: translateY` to keep native acceleration. What do you think?

sankhadeeproy007:
@stackerine Yes, that's on our roadmap. However as there are a few breaking changes with the latest react-native releases. We had to release our next version ASAP, hence this has been pushed to the next release.

stackerine:
Clear... Thank!

PR #2827, opened 2019-07-29
Fixes #2825

### LINKED ISSUE #2825: Cannot access to row's index in renderRow of List
This commit https://github.com/GeekyAnts/NativeBase/commit/a867e33f3565cd082cb2cf2df66636b743222afc#diff-adb80937e3be06283394efe72848a734
changes the behavior of the renderRow of List. 
Then we can no longer access to the index of each row.
It's ok to change the behavior but we should access to the index anyway.

PlabanJr:


src/basic/List.js
```diff
@@ -11,10 +11,14 @@ class List extends Component {
     if (dataArray) {
       return (
         <FlatList
+          {...this.props}
```
sankhadeeproy007: Please add this line at the end.

PlabanJr: Done 👍

PR #2840, opened 2019-08-01
TypeError fix

sayem314:
Fixes: TypeError: undefined is not an object (evaluating 'this._panResponder.panHandlers')


alejandroclvi:
how does moving that piece of code from the didmount method to the constructor help you? I gave it a try, but I'm still getting the same error.

PR #2869, opened 2019-08-12
Add renderEmpty to index.d.ts

PlabanJr:


index.d.ts
```diff
@@ -91,6 +91,7 @@ declare module "native-base" {
              * Takes a data entry from the data source and should return a renderable component to be rendered as the row.
              */
 			renderItem?: Function;
+			renderEmpty?: Function;
```
smigfu: Could we also add `renderBottom`? If you have a transparent card background you end up with the bottom shining through the top card. You can work around this by using `renderBottom` and returning an empty `<View />`

PR #2931, opened 2019-09-17
Adding icon types to action sheets

jramalho:
Fix to ActionSheets support iconTypes with the iconComponent. Just need to pass the prop iconType, otherwise will pick the FontAwesome Type.


hanykumar:
Hi @jramalho , can you just provide minimal code production for the same PR with screenshot output?

Issue #743, opened 2017-04-14
Deck Swiper Not Update when change datasource


huyta:
```
  constructor(props) {
    super(props);
      this.state = {
        dataSource: ['Init test'],
      };
    }
  
```

```
  render() {
        return (
            <Container>
                <StatusBar hidden={true} />
                <Header/>

                <Content>
                <Button success block onPress={() => this.setState({dataSource: ['Expect change but no thing happen']})}/>
                <DeckSwiper
                   ref= {(ref) => { this.desk = ref;}}
                   dataSource={this.state.dataSource}
                   renderItem={(item) => <Text>{item}</Text>}/>

                 </Content>
              </Container>);
    }
```

```

**Comment by gustavoadt** [comment]
I had the same issue but was able to workaround it by checking if the length of the array was greater than 1 before the return:

render() {
        if (this.state.myArray.length > 1) {
            return (

**Comment by SupriyaKalghatgi** [comment]
@huyta The length of your array, is it > 1?

**Comment by SupriyaKalghatgi** [comment]
Closing this due to inactivity

**Comment by angryanatole** [comment]
@SupriyaKalghatgi I have exactly the same issue when I'm trying to change my cards with data from async request. Length of my array is >1.

**Comment by SupriyaKalghatgi** [comment]
@angryanatole Thought the length of your array is > 1, you said `initial datasource is an empty array in my state`.
In any case for Deck Swipe to render, the length of array should be > 1

**Comment by angryanatole** [comment]
@SupriyaKalghatgi got it, thank you for your help!

**Comment by gustavoadt** [comment]
I think the problem is with the DeckSwiper render.
When I click a button to set a new state to my state array (applied to DeckSwiper datasource), it does not render again, not changing the cards automatically. But if I swipe to left or right, it updates the cards...
Maybe it renders, but it does not update the cards by it's own.

**Comment by 123lk** [comment]
Hi @gustavoadt did you find a solution to this?  I'm having the same issue as you.

**Comment by abhishek0058** [comment]
+1 
I only render my DeckSwiper when my data.length is >1 but still it is not refreshing

**Comment by SupriyaKalghatgi** [comment]
@gustavoadt @123lk @abhishek0058 Check this in NativeBase KitchenSink

**Comment by abhishek0058** [comment]
To check What ??

**Comment by luigimassa** [comment]
Hi
I copied the source of `DeckSwiper.js` in my project 
in method  `componentDidUpdate`, I added the check 
```
....
[...]

PR #2934, opened 2019-09-19
Deck Swiper Not Update when change datasource #743

luigimassa:
Update of Decks is only with array length difference. 
Now the update is also with differents object in index 1


shivrajkumar:
A similar PR has already been merged and the issue seems to fixed now.
Please let us know if you still face any issue regarding this.

PR #2935, opened 2019-09-19
Fix crashing issue on Tabs when scrollView ref is not defined

marcioferlan:
🚨  **URGENT!** 🚨

This PR fixes a crashing issue on `src/basic/Tabs/index.js` on `componentDidMount()` function when `scrollView` ref is not defined.

**NOTE:** This is a blocking issue in one of our projects so we are referencing the dependency from my forked repository until this PR gets merged.

**Error:**
> SIGABRT: Unhandled JS Exception: null is not an object (evaluating 'this.scrollView.scrollTo')
> Location: src/basic/Tabs/index.js:80


sankhadeeproy007:
Thanks for the fix, this was fixed in a later PR which has already been merged!

101abhishekgoel:
When it will be uploaded to npm

PR #2946, opened 2019-09-26
Allow override header style

danieloprado:


gabrielmicko:
I am using the latest version of native-base, but in the source code in node_modules I see this problem as well. This should be solved.

PR #2973, opened 2019-10-28
Added swipe to dismiss on `ToastContainer`

rsmelo92:
Currently `Toast` only is dismissed by clicking on a button, what if we need to show a toast without a button, that stays a long time but user can dismiss after the user sees the message?

This implements the solution by using swipe to dismiss only using react native animated


rsmelo92:
Sdds Merge

PR #2979, opened 2019-11-05
Added underlayColor props to ListItem

matgentili:
- Added underlayColor props to ListItem

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -366,6 +366,7 @@ declare module "native-base" {
              */
 			androidRippleColor?: string;
 			touchableHighlightStyle?: RnViewStyleProp;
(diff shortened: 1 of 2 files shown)


samipshah100:
@matgentili this doesnt seem to be working. please see: https://github.com/GeekyAnts/NativeBase/issues/3028 

thanks

annjawn:
This still doesn't work. I had to override the `listBtnUnderlayColor` variable in `platform.js` but that just globally overrides the ListItem touchable feedback color. There seems to be no way of defining a `ListItems` touchable highlight color. I have tried the `touchableHighlightStyle` props but that is useless as well since it simply sets the background color of the `ListItem`.

PR #2997, opened 2019-11-14
fix(item): Foating label with initial value

### LINKED ISSUE #2895: [IOS] Floating label has incorrect styles when rendering an input with an initial value
- [X]   Check latest documentation: https://docs.nativebase.io/
- [X]   Check for existing open/closed issues for a possible duplicate before creating a new issue: https://github.com/GeekyAnts/NativeBase/issues
- [X]   Use the latest NativeBase release: https://github.com/GeekyAnts/NativeBase/releases
- [X]   Check examples from NativeBase KitchenSink https://github.com/GeekyAnts/NativeBase-KitchenSink

## Issue Description
When `<Input />` is initialized with a value, the floating label starts with styles associated as an empty value. Upon clicking on the input, the floating label 'flicks' to its correct styling.
It seems this is only happening on IOS, on Android it seems fine.

This seems to have started happening on version `2.12.2` and is still happening on most recent version `2.13.5`. on version `2.12.1` the floating label acts as expected

sharing a minimal example snack URL exampling the problem:
https://snack.expo.io/rkWtoq64H
(if you run the ios simulator within snack example, the bug only happens if you change `The label` to anything else after the simulator loaded the app)

### Expected behaviour
The floating label starts with correct styles

### Actual behaviour
The floating label starts with incorrect styles

### Steps to reproduce
```
import * as React from 'react';
import { StyleSheet } from 'react-native';
import {
  Item,
  Label,
  Input,
  Root,
  View,
} from 'native-base';

export default class App extends React.Component {
  render() {
    return (
      <Root>
        <View style={styles.container}>
          <Item floatingLabel>
            <Label>
              The label
            </Label>
            <Input
              value="hello"
            />
          </Item>
        </View>
      </Root>
    );
  }
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
});
```

Seems to be only happening on IOS

### Any other additional info which would help us debug the issue quicker.
--

aroucha-miguel:
Fixes #2895

### DIFF
--- a/src/basic/Item.js
+++ b/src/basic/Item.js
@@ -34,8 +34,11 @@ class Item extends Component {
   componentDidMount() {
     if (this.props.floatingLabel) {
       if (this.inputProps && this.inputProps.value) {
-        this.setState({ isFocused: true });
+        const effect = () => {
+          this.setState({ isFocused: true });
+        };
         this.floatUp(-16);
+        effect();
       }
       if (this.inputProps && this.inputProps.getRef)
         this.inputProps.getRef(this._inputRef);
--- a/src/basic/Item.js
+++ b/src/basic/Item.js
@@ -25,7 +25,7 @@ class Item extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      isFocussed: true,
+      isFocused: false,
       text: '',
       topAnim: new Animated.Value(18),
       opacAnim: new Animated.Value(1)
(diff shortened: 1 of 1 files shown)


sankhadeeproy007:
Thanks for the fix!

PR #2887, opened 2019-08-20
Small optimizations

cristianoccazinsp:
Replace some of the mostly used components with PureComponent, add some code tweaks to reduce the component's overhead, and use RN's style merger instead of manually reducing the list.

Please review! I've tested it on a Moto G5 and although the improvements are minimal when the components are re-rendered due to props change or mount, it can be noticeable on re-renders.

Wish some of the styled buttons overhead could be reduced even further.

Issue #1934, opened 2018-05-25
I can't make Toast stay on top of the modal


lyseiha:

SupriyaKalghatgi:
#985

Issue #985, opened 2017-06-26
[Bug] Toast is not working within a Modal


JLLLinn:
## react-native, react and native-base version
RN: 0.45.1
React:16.0.0-alpha.12
native-base: 2.1.5
## Expected behaviour
Toast to be shown in react native Modal
## Actual behaviour
The toast is not shown in react native Modal
## Steps to reproduce (code snippet or screenshot)
Do Toast.show() in a Modal
## Is the bug present in both ios and android or in any one of them?
No that I know of. This was working with RN on 0.44 and native-base on a lower version (I can't remember exactly which one, either 2.1.4 or 2.1.3)

shivrajkumar:
@JLLLinn Check the latest of NativeBase i.e, v2.2.0

laxgoalie392:
Was this verified that it was fixed? I think I'm experiencing it as well. My issue is that the Modal is covering the Toast so it isn't visible.

I'm on RN 0.46.4 and native base 2.2.1 and using the ios emulator.

Thanks!

vovanmozg:
Same problem. I'm on RN 0.46.1 and native base 2.2.1 and using the ios emulator.

mdebo:
Exactly the same problem using RN 0.47 and NativeBase 2.2.1.
Any news about it?
thanks!

gianpaj:
The **Toast** still appears to me behind the **Modal** (from React Native):

![toast behind modal](https://media.giphy.com/media/xUOxfoxIwi2nBcyc00/giphy.gif)

the code is basically this:
```js
import React from 'react';
import {
  Modal,
  Toast,
  Text,
  View,
} from 'react-native';

export default class LoginScreen extends React.Component<Props, State> {
  state = {
    modalVisible: false,
  }

  onResetPassword() {
    Toast.show({
      text: message,
      duration: 2000,
      position: "top",
      textStyle: { textAlign: "center" },
    });
  }    

  render() {
    return (
      <View>
        <Modal
          animationType="slide"
          transparent={false}
          visible={this.state.modalVisible}
          >
          <View>...</View>
        </Modal>
      </View>
    )
  }
}
```

Any suggestions?

I see there's something called [zIndex](http://facebook.github.io/react-native/docs/layout-props.html#zindex) as Layout Props - React Native

Using:
```
$ yarn list --depth=0 | grep 'native-base@\| react-native@'
├─ native-base@2.3.3
├─ react-native@0.50.3
```

[...]

PR #3018, opened 2019-12-09
Multiple roots

cristianoccazinsp:
fix toast and actionsheet with multiple root elements (e.g., for modals). Should fix/help with https://github.com/GeekyAnts/NativeBase/issues/1934 and https://github.com/GeekyAnts/NativeBase/issues/985 when multiple Root elements are used in order to show toast/actionsheets on a modal.

Approach: keep a static list of root instances on mount/unmount, and always use the last one that was mounted. This should be consistent with modals mount order.

Caveat: there's now a cyclic reference between Root and Toast/Actionsheet containers.

Possible improvements: Use Contexts; however, we do not have context access from static methods, so this is quite harder to get working.

Update: PR carried changes from my previous (unmerged) PR that added some performance improvements to Text, Button, and Icons (https://github.com/GeekyAnts/NativeBase/pull/2887)


patricktan98:
May I know why the merging is blocked?
Problem still persisted.

PR #3066, opened 2020-02-12
Tabs not using react-timer-mixin

dgarciamuria:
Currently, `Tabs` implementation applies `react-timer-mixin` but the `setTimeout` in `componentDidMount` is not using it since it should be `this.setTimeout`. In consequence, there might be cases where that callback gets fired and when the component is already unmounted. I hit that error my self, were `scrollView` was null and it was trying to call `.scrollTo`

Cheers!


FDiskas:
But the bad thing is that this module https://calendar.perfplanet.com/2019/the-unseen-performance-costs-of-css-in-js-in-react-apps/ is deprecated

FDiskas:
if you want to remove mixin component should be refactored to react class and timer should be assigned to private variable and on unmount timeout should be cleared

sankhadeeproy007:
@dgarciamuria @FDiskas What's the status with this one? Can this be merged?

FDiskas:
@sankhadeeproy007 it can be merged. Later it should be refactored and unmaintained mixin removed.

dgarciamuria:
@FDiskas I refactored and removed the mixin dep. Wasn't sure if needed to run the `prepublish` script or not, but that was the only way to update the `dist` directory. Let me know if that's correct since now IconNB is in conflict.

Thanks!

FDiskas:
I suppose the dist directory should be touched by the owners and only before releasing new version. So in this case only files in src directory should be changed.

dgarciamuria:
@FDiskas cool. I just reverted that piece

FDiskas:
There is something strange is going on in `yarn.lock` file.

dgarciamuria:
@FDiskas mmm, it seems that `yarn.lock` has been out of sync for a while, so after running `yarn remove react-timer-mixin` it got re generated. Same thing happens when running `yarn install`. But anyways I reverted `yarn.lock`

FDiskas:
@sankhadeeproy007 ready steady go :)

Issue #2962, opened 2019-10-10
button auto full width 


kin-for-test:
expect 
![fromweb](https://user-images.githubusercontent.com/23132046/66554522-b9ace300-eb7f-11e9-863e-1859f7dc60be.png)

but the view in android vm like 
![fromandroid](https://user-images.githubusercontent.com/23132046/66554564-cfbaa380-eb7f-11e9-8884-bb4bf8137d96.png)

code 
 

```
import React, { Component } from 'react';
import { Container, Header, Content, Button, Text } from 'native-base';
export default class ButtonThemeExample extends Component {
  render() {
    return (
      <Container>
        <Header />
        <Content>
          <Button light><Text> Light </Text></Button>
          <Button primary><Text> Primary </Text></Button>
          <Button success><Text> Success </Text></Button>
          <Button info><Text> Info </Text></Button>
          <Button warning><Text> Warning </Text></Button>
          <Button danger><Text> Danger </Text></Button>
          <Button dark><Text> Dark </Text></Button>
        </Content>
      </Container>
    );
  }
}
```

i try full={false}  but still not work

Mohammad-Khalid23:
**full={false}** will not work, because when you pass full = {true} then it will apply full button style , other wise in default, button will get full width of the parent container because it is using flex for button style.

you have to give width to the buttons.

kin-for-test:
OH! i see , i thought the button have a default width , because the button show  in docs is not full width

gsouf:
I actually found the doc to be terrible, lot of actual examples do not match the result ;(

mkamyszek:
What is the solution to get it to look like the docs?

ahmadsyamim:
```
<Content>
<View style={{
flex:0
}}>
```

should be fine i think.

cleybertandre:
I've solved it adding a 'alignSelf' style property on the Button:

```
<Button primary style={{alignSelf: 'flex-start'}}><Text> Primary </Text></Button>
```

sorxrob:
> I've solved it adding a 'alignSelf' style property on the Button:
> 
> ```
> <Button primary style={{alignSelf: 'flex-start'}}><Text> Primary </Text></Button>
> ```

apply this won't let you use block and full properties

hanykumar:
Hi @sorxrob, Button doesn't lose any styles using
`<Button primary style={{alignSelf: 'flex-start'}}><Text> Primary </Text></Button>`.
[...]

PR #3094, opened 2020-03-13
bug fix 2962

hanykumar:
Hi, Issue: #2962, Buttons fixed.
Screenshot 2020-03-13 at 5 38 20 PM

### DIFF
--- a/src/theme/components/Button.js
+++ b/src/theme/components/Button.js
@@ -380,7 +380,8 @@ export default (variables /* : * */ = variable) => {
     shadowOpacity: platformStyle === PLATFORM.MATERIAL ? 0.2 : undefined,
     shadowRadius: platformStyle === PLATFORM.MATERIAL ? 1.2 : undefined,
     alignItems: 'center',
-    justifyContent: 'space-between'
+    justifyContent: 'space-between',
+    alignSelf: 'flex-start'
   };
   return buttonTheme;
 };


src/theme/components/Button.js
```diff
@@ -380,7 +380,8 @@ export default (variables /* : * */ = variable) => {
     shadowOpacity: platformStyle === PLATFORM.MATERIAL ? 0.2 : undefined,
     shadowRadius: platformStyle === PLATFORM.MATERIAL ? 1.2 : undefined,
     alignItems: 'center',
-    justifyContent: 'space-between'
+    justifyContent: 'space-between',
+    alignSelf: 'flex-start'
```
sankhadeeproy007: Can we do it without using alignSelf? The alignment of the button should be left to the user.

hanykumar: I think `alignSelf` is fine. I don't find any other solution. The alignment of the button is left to the user. Isn't it?

sankhadeeproy007: I mean the user might want to align it to the center or to the right. So I think it should be the user's choice, can we use `alignSelf: 'auto'` and see if that works?

hanykumar: `alignSelf: 'auto' `will set width to full. But User always has a choice for `alignSelf: 'center'` whether we set `alignSelf: 'flex-start'` by default or not.

Issue #2945, opened 2019-09-26
Warning: Received `false` for a non-boolean attribute `uppercase`


arsamigullin:
Hi. Getting this warning in a web browser console

bundle.js:144396 Warning: Received `false` for a non-boolean attribute `uppercase`.

If you want to write it to the DOM, pass a string instead: uppercase="false" or uppercase={value.toString()}.

If you used to conditionally omit it with uppercase={condition && value}, pass uppercase={condition ? value : undefined} instead.
    in div (created by Text)
    in Text (at Text.js:26)
    in Text (at connectStyle.js:392)
    in Styled(Text) (at SettingsScreen.js:19)
    in div (created by View)
    in View (at Left.js:10)
    in Left (at connectStyle.js:392)
    in Styled(Left) (at SettingsScreen.js:18)
    in div (created by View)
    in View (at ListItem.js:39)
    in div (created by View)
    in View (created by TouchableHighlight)
    in TouchableHighlight (at ListItem.js:31)
    in ListItem (at connectStyle.js:392)
    in Styled(ListItem) (at SettingsScreen.js:17)
    in div (created by View)
    in View (at List.js:25)
    in List (at connectStyle.js:392)
    in Styled(List) (at SettingsScreen.js:15)
    in div (created by View)
    in View (created by ScrollView)
    in div (created by View)
    in View (created by ScrollViewBase)
    in ScrollViewBase (created by ScrollView)
    in ScrollView (at KeyboardAwareHOC.js:517)
    in KeyboardAwareScrollView (at Content.js:66)
    in Content (at connectStyle.js:392)
    in Styled(Content) (at SettingsScreen.js:14)
    in div (created by View)
    in View (at Container.js:12)
    in Container (at connectStyle.js:392)
    in Styled(Container) (at SettingsScreen.js:13)
    in SettingsScreen (at SceneView.js:9)
    in SceneView (at StackViewLayout.tsx:898)
    in div (created by View)
    in View (at StackViewLayout.tsx:897)
    in div (created by View)
    in View (at StackViewLayout.tsx:896)
    in div (created by View)
    in View (created by AnimatedComponent)
    in AnimatedComponent (at StackViewCard.tsx:106)
    in div (created by View)
    in View (created by AnimatedComponent)
    in AnimatedComponent (at screens.web.js:105)
    in Screen (at StackViewCard.tsx:93)
    in Card (at createPointerEventsContainer.tsx:95)
[...]

PR #3111, opened 2020-03-31
fix(): warning uppercase a text component

juandav:
This happens since they are sending invalid properties with this.props

```
Warning: Received `true` for a non-boolean attribute `uppercase`.

If you want to write it to the DOM, pass a string instead: uppercase="true" or uppercase={value.toString()}.
```


juandav:
#2945

hube12:
Could this be accepted?

PR #3136, opened 2020-04-20
added useNativeDriver:false to Animated.timing|spring|event

dansokol:
fixes bug with react native version .62+ that now requires useNativeDriver to be specified:
https://github.com/facebook/react-native/commit/5876052615f4858ed5fc32fa3da9b64695974238


sankhadeeproy007:
This PR can have performance impact on a few animations. We need to check what properties are not supported by `nativeDriver` and set them false. 
https://reactnative.dev/docs/animations#caveats

levibuzolic:
The default was already `false` so this wont have any performance regressions. A separate task would be to go through and find animations that could be switched to native only.

sankhadeeproy007:
Okay, makes sense. If it only adds false non-existing properties.
@dansokol Can you resolve the conflicts so that we can merge it?

dansokol:
@sankhadeeproy007 resolved conflicts.

craftzdog:
Great fix. Could you roll it out?

n3rd253:
@sankhadeeproy007 

Was a change rolled out here?

PR #3163, opened 2020-05-27
Picker has been extracted from react-native core

Preeternal:
The [issue](https://github.com/GeekyAnts/NativeBase/issues/3114)
Could you please to review?


Preeternal:
replaced community-picker into peerDependencies

lifenautjoe:
Can someone please merge and release this? Blocking as it prevents app instantiation with 
```
 Invariant Violation: Tried to register two views with the same name AndroidDropdownPicker
```

alexandrtovmach:
@Preeternal I'm not sure that's moving required dependency to `peerDependencies` is good idea

#3230 it's duplicate, but it used "normal" `dependencies`, so my vote for that

Preeternal:
> @Preeternal I'm not sure that's moving required dependency to `peerDependencies` is good idea
> 
> #3230 it's duplicate, but it used "normal" `dependencies`, so my vote for that
But someone probably will want to use community picker and native-base in own project same time. It will return error described by @lifenautjoe

Issue #2238, opened 2018-08-24
[ts] "native base has no exported module connectStyle"


Eddie-CooRo:
## node, npm, react-native, react and native-base version, expo version if used
native-base@2.7.2
## Expected behaviour
Importing "connectStyle" from nativebase
## Steps to reproduce
```npm i native-base@latest```
```import {connectStyle} from 'native-base'```
## Actual behaviour
"connectStyle" HOC should get exported inside index.d.ts

vinay0458:
temporary solution for writing jest test cases, 
ignore nativebase like this in jest
"transformIgnorePatterns": [
      "/node_modules/(?!react-native|react-clone-referenced-element|react-navigation|native-base)"
    ],

cargallo:
Any Solution to this?

AmalMenachery:
@SupriyaKalghatgi The Issue is still open in 2.13.12, Is the fix expected soon ?

PR #3180, opened 2020-07-08
Added exported connectStyle to type definition

lochiego:
Addresses #2238

Type definition was missing an export for 'connectStyle' helper function that's re-exported from native-base-shoutem-theme. Also addressed several linter errors that were caught by pre-commit checks.


komputerwiz:
> @komputerwiz is attempting to deploy a commit to the **Geekyants Team** Team on [Vercel](https://vercel.com).

I was a little confused at first since I did not initiate this.

It looks like the _Vercel_ bot is just using the committer info on the HEAD commit.

PR #3195, opened 2020-07-20
fix(types): add  floatBack for Label

GregoryEsberci:
Add missing type declaration `floatBack`  property for `Label`, used by [Item](https://github.com/GeekyAnts/NativeBase/blob/5c190746544293aa9c434c4831771f7e17043317/src/basic/Item.js#L423)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -648,6 +648,7 @@ declare module "native-base" {
 
 		interface Label extends Testable {
 			style?: RnTextStyleProp;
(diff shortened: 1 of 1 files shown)


suhail135:
**Items with floatingLabel are broken in release mode** 

"native-base": "^2.13.14",
 "react-native": "0.63.0",

KawaljeetSBagga:
@suhail135  any potential fix ??

PR #3205, opened 2020-07-29
Fix wrong package import for typescript definition

ptondereau:


KovaxG:
Hey, since we upgraded to the latest version of typescript we've been getting this when running jest

```
Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain 

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_mod

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePa
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/project/node_modules/@codler/react-native-keyboard-aware-scroll-view/index.js:3
    import listenToKeyboardEvents from './lib/KeyboardAwareHOC'
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime._execModule (node_modules/jest-runtime/build/index.js:1179:56)
      at Object.<anonymous> (node_modules/native-base/src/basic/Content.js:5:1)
```
The package might need some work

ptondereau:
> Hey, since we upgraded to the latest version of typescript we've been getting this when running jest
> 
> 
> 
> ```
> 
> Jest encountered an unexpected token
> 
> 
> 
>     This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain 
> 
> 
> 
>     By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_mod
> 
> 
> 
>     Here's what you can do:
> 
>      • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePa
> 
>      • If you need a custom transformation specify a "transform" option in your config.
> 
>      • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the
> 
> 
> 
>     You'll find more details and examples of these config options in the docs:
> 
>     https://jestjs.io/docs/en/configuration.html
> 
> 
> 
>     Details:
> 
> 
> 
>     /home/project/node_modules/@codler/react-native-keyboard-aware-scroll-view/index.js:3
> 
>     import listenToKeyboardEvents from './lib/KeyboardAwareHOC'
> 
>     ^^^^^^
> 
> 
> 
>     SyntaxError: Cannot use import statement outside a module
> 
> 
> 
>       at Runtime._execModule (node_modules/jest-runtime/build/index.js:1179:56)
> 
>       at Object.<anonymous> (node_modules/native-base/src/basic/Content.js:5:1)
> 
> ```
> 
> The package might need some work

Add package name to the transformIgnorePattern key in your jest config

KovaxG:
You are right, thank you for your suggestion :)

davidgovea:
Hi there!

I just added a PR that addresses this issue as well - whoops! Looked through the issues before adding, but not the PRs.
#3209 #3210 

That PR takes a different approach -- moving back to the upstream library. Since most will probably already have `react-native.*` entry in their jest `transformIgnorePatterns`, that could be another consideration when deciding whether to stay on the forked library or go back to the (now fixed) upstream version.

ptondereau:
> Hi there!
> 
> 
> 
> I just added a PR that addresses this issue as well - whoops! Looked through the issues before adding, but not the PRs.
> 
> #3209 #3210 
> 
> 
> 
> That PR takes a different approach -- moving back to the upstream library. Since most will probably already have `react-native.*` entry in their jest `transformIgnorePatterns`, that could be another consideration when deciding whether to stay on the forked library or go back to the (now fixed) upstream version.

Feel free to close This PR. But it solves definition autocomplete with typescript...
E.g: 'enableOnAndroid' is not available anymore...

Issue #3209, opened 2020-08-03
Broken types for Content component


davidgovea:
## I have gone through these following points

- [x] Check latest documentation: https://docs.nativebase.io/
- [x] Check for existing open/closed issues for a possible duplicate before creating a new issue: https://github.com/GeekyAnts/NativeBase/issues
- [x] Use the latest NativeBase release: https://github.com/GeekyAnts/NativeBase/releases
- [x] Check examples from NativeBase KitchenSink https://github.com/GeekyAnts/NativeBase-KitchenSink
- [x] For discussion purpose make use of NativeBase Slack: http://slack.nativebase.io/
- [x] For queries related to theme, check [Theme Variables](https://docs.nativebase.io/docs/ThemeVariables.html) from Docs and live NativeBase Theme Editor http://nativebase.io/customizer/

## Issue Description

PR #3185 broke the typescript definitions for the `Content` component.

@codler [changed the imported `react-native-keyboard-aware-scroll-view`](https://github.com/GeekyAnts/NativeBase/pull/3185/files#diff-73890935d8b7cec43384f2b26428017bR5) to his own fork, but did not update the version imported in the `index.d.ts` file: https://github.com/GeekyAnts/NativeBase/blob/master/index.d.ts#L4

This leads to type errors like the following: 
`Property 'bounces' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Content> & Readonly<Content> & Readonly<{ children?: ReactNode; }>'.`
unless the consuming project manually adds the upstream version of react-native-keyboard-aware-scroll-view to its dependencies as well (meaning, installing both the fork and upstream packages side-by-side).

Since the [upstream `react-native-keyboard-aware-scroll-view`]() has [fixed the issue](https://github.com/APSL/react-native-keyboard-aware-scroll-view/releases/tag/v0.9.2), **I suggest that native-base goes back to the upstream library.**

---
Since `patch-package` cannot be used to patch dependencies in `package.json`, here is a quick patchfile to fix the issue by importing the forked library in the typedef:

**patches/native-base+2.13.13.patch**
```
diff --git a/node_modules/native-base/index.d.ts b/node_modules/native-base/index.d.ts
index 73c40e9..8c0b92a 100644
--- a/node_modules/native-base/index.d.ts
+++ b/node_modules/native-base/index.d.ts
@@ -1,7 +1,7 @@
 declare module "native-base" {
 	import * as React from "react";
[...]

PR #3210, opened 2020-08-03
Fix <Content> typescript definition & Revert use of forked library

davidgovea:
Motivation/error explained in Issue #3209 

This PR switches `native-base` back to the upstream version of [`react-native-keyboard-aware-scroll-view`](https://github.com/APSL/react-native-keyboard-aware-scroll-view) in order to fix typescript errors on the <Content> component.  It was nice of @codler to provide a quick fix for RN 0.63, but since the fix was incompletely implemented, and the [upstream repository has active maintenance](https://github.com/APSL/react-native-keyboard-aware-scroll-view/issues/440#issuecomment-659358713), I think it's appropriate to switch back to the main library.

Thanks for reading!

---

TO DISCUSS:
NativeBase uses a combination of locked, semver-compat ("^version"), and minor-version ("~version") specifications for dependencies. I used a locked version to match the previous style, but perhaps the version of `react-native-keyboard-aware-scroll-view` could be minor-version unlocked? 
It seems that more minor deprecations & changes may be coming (see here: https://github.com/APSL/react-native-keyboard-aware-scroll-view/issues/440), so perhaps unlocking the version could allow consuming apps to receive the minor patches without requiring a NativeBase update.

Let me know what you think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
   "license": "Apache-2.0",
   "private": false,
   "dependencies": {
-    "@codler/react-native-keyboard-aware-scroll-view": "1.0.0",
     "blueimp-md5": "^2.5.0",
     "clamp": "^1.0.1",
     "color": "~3.1.2",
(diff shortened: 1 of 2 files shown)


shivrajkumar:
@davidgovea A simialr PR with the fix has already been merged

Issue #3237, opened 2020-09-09
Transparent buttons in the header does not have visual feedback on Android.


zoolyka:
Using transparent Button in the Header component Left or Right side results in no visual feedback when pressing them on Android. On iOS it is clearly visible.

NativeBase version: 2.13.14 (recent)

There was an issue created for this [2595](https://github.com/GeekyAnts/NativeBase/issues/2595), but it was closed without any fix.

Thanks.

      <Container>
        <Header>
          <Left>
            <Button transparent>
              <Icon name='arrow-back' />
            </Button>
          </Left>
          <Body>
            <Title>Header</Title>
          </Body>
          <Right>
            <Button transparent>
              <Icon name='menu' />
            </Button>
          </Right>
        </Header>
      </Container>

PR #3246, opened 2020-09-16
fix: Transparent buttons in the header now have feedback in android

hanykumar:
#3237


cyrus25 (approved):

geekabhinav:
PR seems to add unnecessary commas and round braces which is not related to the change. @cyrus25 Are you sure this is ready to be merged?

Issue #3140, opened 2020-04-26
Transparent header stopped working


Agilulfo1820:
## I have gone through these following points

- [x] Check latest documentation: https://docs.nativebase.io/
- [x] Check for existing open/closed issues for a possible duplicate before creating a new issue: https://github.com/GeekyAnts/NativeBase/issues
- [x] Use the latest NativeBase release: https://github.com/GeekyAnts/NativeBase/releases
- [x] For queries related to theme, check [Theme Variables](https://docs.nativebase.io/docs/ThemeVariables.html) from Docs and live NativeBase Theme Editor http://nativebase.io/customizer/

#### node, npm, react-native, react and native-base version, expo version if used, xcode version
"react": "16.9.0",
"expo": "^37.0.0",
"react-navigation": "^4.0.10",
"native-base": "2.13.8",
#### Expected behaviour
When using `<Header transparent/>` the  android status bar should have the color of the header with a bit of opacity.
#### Actual behaviour
The Header is transparent but the android status bar takes the default color theme. Now when going to a screen without the `<Header/>` component the android status bar has a black color and a margin bottom (which wasn't there before).
#### Steps to reproduce
I created a repo with expo, react navigation and native base to demonstrate the issue.
https://github.com/Agilulfo1820/mvce-header-bug

#### Is the bug present in both iOS and Android or in any one of them?
The bug is present in both iOS and Android.
#### Any other additional info which would help us debug the issue quicker.
This bug wasn't present since 3 weeks ago (after 5 months of developing). Don't know if it's related to expo or react navigation.

Here are a few screenshots:
How broken header presents:
![78924529-4f2b2000-7a9a-11ea-825b-517e55cdd94a](https://user-images.githubusercontent.com/6793051/80314524-52404300-87f2-11ea-96d2-d5fe17d04a12.jpg)

Before entering screen with the broken header:
![78924533-505c4d00-7a9a-11ea-882d-3ab498775edc](https://user-images.githubusercontent.com/6793051/80314526-53717000-87f2-11ea-9ba8-d452845afeb5.jpg)

Returning to this screen after been to the screen with broken header:

![78924531-4fc3b680-7a9a-11ea-9cbb-49c6735582f1](https://user-images.githubusercontent.com/6793051/80314525-53717000-87f2-11ea-90ae-949eb8acb87e.jpg)

swaraasolutions:
Hi Guys any update on this, i am also facing similar issue

[...]

PR #3248, opened 2020-09-16
fix: android statusbarcolor for transparent header

hanykumar:
#3140 
android statusbarcolor for transparent header fixed.
Brefore:
Screenshot 2020-09-16 at 1 27 14 PM

After:
Screenshot 2020-09-16 at 1 26 59 PM


cyrus25 (commented):
not working for android

Issue #3268, opened 2020-10-05
Textarea Typescript fields should be optional


Aryk:
I wrote a comment on this merge:

https://github.com/GeekyAnts/NativeBase/pull/2756

I'm using 2.13.8 (2.13.14 also has this issue). For `Textarea`, type definitions are currently:

```ts
		interface Textarea extends ReactNative.TextInputProps, Testable {
			rowSpan: number;
			bordered: boolean;
			underline: boolean;
			/**
             * Disables inputting data.
             */
			disabled?: boolean;
		}
```

But should be:

```ts
		interface Textarea extends ReactNative.TextInputProps, Testable {
			rowSpan?: number;
			bordered?: boolean;
			underline?: boolean;
			/**
             * Disables inputting data.
             */
			disabled?: boolean;
		}
```

They should not be **required** to be entered...am I right?

ARKEOLOGIST:
I'm not sure about rowspan, but I don't think bordered and underline are necessary.

Aryk:
Thanks for making the PR, pretty sure rowspan is not required as my app is performing fine with it.

ARKEOLOGIST:
You mean without it?

PR #3270, opened 2020-10-07
Made props for Textarea optional

ARKEOLOGIST:
The props for Textarea should be optional and not mandatory.


ARKEOLOGIST:
This should fix #3268

Issue #3017, opened 2019-12-06
Open accordion items upwards


sachinshettigar:
Feature Request

Can we have a option to open the accordion items to be expanded upwards, instead of the default downwards. Expecting a toggle props eg: headerOnTop : boolean , for Accordion which will by default have current behavior, but we can update this value to expand accordion upwards too.

PR #3286, opened 2020-10-31
enable the ability to open accordion content on top passing a boolean (issue#3017)

NihaalAftab:
This PR contains the implementation of a `feature request` (issue#3017) which is to enables the functionality to open accordion content on top rather than bottom by passing a boolean prop.
Thus now by passing `showContentOnTop={true}`  will render content on top.

For reference:
#3017


cyrus25 (approved):

shivrajkumar:
@NihaalAftab It would be great if you could resolve the conflicts for us to take it forwardd

NihaalAftab:
Sure, on it..

just discovered while resolving conflicts, `DefaultHeader` is no more receiving `title` prop now, is that intentional? since it is merged in master already.

https://github.com/GeekyAnts/NativeBase/blob/master/src/basic/Accordion.js#L143

PR #3318, opened 2020-12-18
fix: fixed web compatiblity for sliders

MD-REHMAN:


src/components/primitives/Slider/SliderThumb.tsx
```diff
@@ -19,17 +19,20 @@ const SliderThumb = ({ children, ...props }: ISliderProps) => {
     SliderThumb: {
       position: 'absolute',
       display: 'flex',
-      left: sliderOffset,
+      // @ts-ignore - cheking for undefined in the context itself.
```
theankurkedia, line 22: It does not work like that. You will have to add checks here too. Please avoid @ts-ignore unless it is unavoidable.

src/theme/components/slider.ts
```diff
@@ -8,13 +8,15 @@ const baseStyle = (props: Record<string, any>) => {
       `${simplifiedColorScheme}.200`
     )(props),
     trackColor: mode('gray.200', 'gray.700')(props),
+    sliderSizeOnWeb: 200,
```
theankurkedia, line 11: If you adding something in theme, please add some standard style prop. Ideally should rethink it and calculate it based on size prop only (inside the component itself)

src/components/primitives/Slider/SliderTrack.tsx
```diff
@@ -12,7 +12,7 @@ const SliderTrack = ({ children, ...props }: ISliderProps) => {
     <Box
       position="absolute"
       backgroundColor={trackColor}
-      height={sliderSize}
+      height={`${sliderSize}px`}
```
theankurkedia, line 15: this is fixed now, can remove this. using size in number and px are two different things.

src/theme/components/slider.ts
theankurkedia, line 11: Please remove this.

PR #3321, opened 2020-12-20
Change React Native Picker package to point to correct namespace

armanatz:
NativeBase is currently using the old namespace for `@react-native-community/picker` when it has now been changed to `@react-native-picker/picker`. This PR aims to fix that so that the library references the currently maintained package.


alexandrtovmach (approved):
LGTM!

PR #3326, opened 2020-12-23
feat: useTypeahead hook and Typeahead component

rohits-geekyants:


theankurkedia:
Please make the following changes:
• Move the hook to Typeahead component directory, since it is not a generic hook.
• Merge the examples

Bugs:
• on iOS, if I type in something and the input is empty, it shows all options. Ideally, it should be closed if input is empty unless it is opened from button.
• When I type, if I select something, it does not get selected at once, it takes two clicks.
• When dropdown is opened, I cannot remove the focus from input. Ideally, should be able to de-select and close the dropdown by clicking outside.

src/components/composites/Typeahead/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+    mt,
+    mb,
+    ml,
+    mr,
+    ...props
+  }: ITypeaheadProps,
+  ref: any
+) => {
+  const [inputItems, setInputItems] = React.useState(data);
+  const marginalProps = { m, mt, mb, ml, mr };
```
theankurkedia, line 44: Why only margin props? Why not positioning, padding and all layout props?

theankurkedia: Can use `extractInObject` util function

src/components/composites/Typeahead/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      width="100%"
+      flexDirection="row"
+      onLayout={onLayout}
+      {...marginalProps}
+    >
+      <Box flex={1}>
+        <Input
+          {...props}
+          {...getInputProps(inputValue, onChangeText)}
+          InputRightElement={
```
theankurkedia, line 86: Since this is a prop. Please rename it in Input component using a proper naming convention. Should be `inputRightElement`.

theankurkedia:
Can we make it accessible?

src/components/composites/Typeahead/index.tsx
rohits-geekyants: We have discussed this before since it takes Components as value therefore we have decided to go with InputRightElement.

PR #3339, opened 2021-01-03
Fab: Fix web compatibility

hsjoberg:
This pull request fixes web compatibility with the Fab component.
Right now, when running on web, the component tries to load in `<TouchableNativeFeedback>`, which will throw an exception on web as it's not available.

Cheers


hsjoberg:
Nevermind my PR looks faulty -- it seems like this was already addressed recently.
https://github.com/GeekyAnts/NativeBase/commit/f62104de2b26792673f32c60a4dbc03781a96251

PR #3351, opened 2021-01-07
fix: added themeing and dark mode support

MD-REHMAN:


src/theme/components/app-bar.ts
```diff
@@ -0,0 +1,19 @@
+import { mode, getColorScheme } from './../tools';
+
+const baseStyle = (props: Record<string, any>) => {
+  let colorScheme = getColorScheme(props);
+  return {
+    bg: mode(`${colorScheme}.500`, `${colorScheme}.300`)(props),
+    color: mode('gray.800', 'gray.50')(props),
+    px: 2,
+    statusBarHeight: 2,
```
theankurkedia, line 9: Would this go here? This is not any prop or style that can directly be applied to any component. We should not add any custom prop to theme file. It should be intuitive. This will go in the component as a constant. But why is it hard-coded to be 2?

src/theme/components/app-bar.ts
```diff
@@ -0,0 +1,19 @@
...
+    height: '56px',
+    space: 2,
```
theankurkedia, line 11: This is not a style, so I think it should not go here.

src/theme/components/app-bar.ts
MD-REHMAN: Thanks for correcting me, I by mistake left it here.

PR #3359, opened 2021-01-10
Run Project

jatinggrwl:
Install the dependencies in root project also to run the project


shivrajkumar:
@jatinggrwl Nativebase is a module to be installed for UI in RN, and not actually an RN application. So those commands don't make sense to include in the docs

PR #3399, opened 2021-01-27
fix: checkbox design

vidhi499:


example/storybook/stories/components/primitives/Checkbox/controlledCheckbox.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Checkbox, CheckboxGroup, Text } from 'native-base';
-import type { ICheckboxValue } from 'src/components/primitives/Checkbox/props';
+import type { ICheckboxValue } from '../../../../../../src/components/primitives/Checkbox/props';
```
theankurkedia, line 3: Instead of importing here, we can import it from native-base. In case it is not exported properly to the root of package, then we should export it there. Example repo is an independent project. All the imports should be like you would do in a normal project.

PR #3401, opened 2021-01-28
Feat/circular progress redesign

makkarMeenu:


src/components/composites/CircularProgress/index.tsx
```diff
@@ -52,8 +54,20 @@ const CircularProgress = ({
   } else {
     sizeProps = { height: size, width: size };
   }
+  // fetching size from theme for passing into style prop
+  const themeHeight = get(theme, 'space.' + sizeProps.height);
```
theankurkedia, line 58: Can use useToken hook for this. https://nativebase.github.io/docs/use-token.

src/components/composites/CircularProgress/index.tsx
```diff
@@ -52,8 +54,20 @@ const CircularProgress = ({
...
+  const themeWidth = get(theme, 'space.' + sizeProps.width);
 
-  let defaultThickness = 4;
+  const styleSize = {
+    height: themeHeight
+      ? parseInt(themeHeight.slice(themeHeight.Length, -2))
```
theankurkedia, line 63: Why are we doing this? Is this in px?

makkarMeenu: Yes, getting string in px

PR #3403, opened 2021-01-28
Feat/modal redesign

makkarMeenu:


src/components/composites/Modal/index.tsx
```diff
@@ -98,8 +97,8 @@ const Modal = (
   const modalChildren = (
     <Box
       {...newProps.modalProps}
-      justifyContent={isCentered ? 'center' : justifyContent}
-      alignItems={isCentered ? 'center' : alignItems}
+      justifyContent={justifyContent ? justifyContent : 'center'}
```
theankurkedia, line 100: Can use coalescing operator. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

PR #3416, opened 2021-02-02
fix: focus web fix, web design and crash fix

MD-REHMAN:
- https://trello.com/c/wUVZWqPO
- https://trello.com/c/Ac7RrKKf


src/components/composites/PinInput/index.tsx
```diff
@@ -10,6 +10,7 @@ import { FormControlContext, IFormControlContext } from '../FormControl';
 import { Platform } from 'react-native';
 import { PinInputContext } from './Context';
 import { themeTools } from '../../../theme';
+import { remove } from 'lodash';
```
theankurkedia, line 13: Should use "lodash/remove", otherwise whole lodash will be added to bundle.

PR #3422, opened 2021-02-03
fix: remover extra border

MD-REHMAN:


src/components/composites/Accordion/AccordionButton.tsx
```diff
@@ -27,6 +28,8 @@ const AccordionButton = (
   const pressHandler = () => {
     isOpen ? onClose && onClose() : onOpen && onOpen();
   };
+  console.log(_expanded);
```
theankurkedia, line 31: Please remove this

PR #3427, opened 2021-02-03
Feat/radio test cases

vidhi499:


package.json
```diff
@@ -150,6 +150,7 @@
     ]
   },
   "dependencies": {
+    "@testing-library/react-native": "^7.1.0",
```
theankurkedia, line 153: We should move this to dev dependencies

vidhi499: fixed

Issue #3413, opened 2021-02-01
Accordion titles are not rendered


Kolyunya:
## I have gone through these following points

- [x] Check latest documentation: https://docs.nativebase.io/
- [x] Check for existing open/closed issues for a possible duplicate before creating a new issue: https://github.com/GeekyAnts/NativeBase/issues
- [x] Use the latest NativeBase release: https://github.com/GeekyAnts/NativeBase/releases
- [x] Check examples from NativeBase KitchenSink https://github.com/GeekyAnts/NativeBase-KitchenSink
- [x] For discussion purpose make use of NativeBase Slack: http://slack.nativebase.io/
- [x] For queries related to theme, check [Theme Variables](https://docs.nativebase.io/docs/ThemeVariables.html) from Docs and live NativeBase Theme Editor http://nativebase.io/customizer/

#### node, npm, react-native, react and native-base version, expo version if used, xcode version
```
» node_modules/.bin/react-native info
info Fetching system and libraries information...
System:
    OS: Linux 5.10 Arch Linux
    CPU: (4) x64 Intel(R) Core(TM) i5-7600 CPU @ 3.50GHz
    Memory: 15.17 GB / 23.43 GB
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 15.7.0 - /usr/bin/node
    Yarn: Not Found
    npm: 6.14.11 - /usr/bin/npm
    Watchman: Not Found
  SDKs:
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Java: 10.0.2 - /usr/bin/javac
    Python: 3.9.1 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: ^17.0.0 => 17.0.1 
    react-native: https://github.com/expo/react-native/archive/sdk-40.0.1.tar.gz => 0.63.2 
    native-base: ^2.15.2 => 2.15.2
  npmGlobalPackages:
    *react-native*: Not Found
```

#### Expected behaviour
Accordion title is rendered.

#### Actual behaviour
Accordion title is not rendered.

#### Steps to reproduce
1. Add an `Accordion` component.
2. See that titles are not rendered.

[Expo.io snack](https://snack.expo.io/@kolyunya/bug-demo---native-base-accordion).

![Screenshot_20210201_203715](https://user-images.githubusercontent.com/2682768/106495904-4c883400-64cd-11eb-8000-cc9f2cc595e1.png)

[...]

PR #3434, opened 2021-02-04
accordion title getting showed

cyrus25:
Screenshot_1612444406
Resolves issue #3413 
The issue occurs only on android devices.


Kolyunya:
> The issue occurs only on android devices.

@cyrus25 thanks for your help! I just want to mention that according to my tests the original issue occurs both on Android and iOS devices.

PR #3456, opened 2021-02-10
Feat/radio aria

vidhi499:


src/components/primitives/Radio/RadioGroup.tsx
```diff
@@ -1,37 +1,35 @@
...
-export const RadioContext = React.createContext<IRadioContext>({
-  onChange: (_value: IRadioValue) => {},
-  name: '',
-});
+export let RadioContext = React.createContext<IRadioContext>(
+  {} as IRadioContext
+);
 
-const RadioGroup = ({ size, colorScheme, ...props }: IRadioGroupProps) => {
+const RadioGroup = ({ size, colorScheme, ...props }: any) => {
```
intergalacticspacehighway, line 13: We shouldn't use `any` here.

package.json
```diff
@@ -91,7 +91,7 @@
       "<rootDir>/lib/"
     ],
     "transformIgnorePatterns": [
-      "node_modules/(?!(react-native|styled-components|expo-asset|expo-constants|@unimodules|react-native-unimodules|expo-font|react-native-svg|@expo/vector-icons|styled-system|react-native-vector-icons|@react-native-aria/interactions|@react-native-aria/button|@react-native-aria/switch|@react-native-aria/toggle|@react-native-aria/utils)/)"
+      "node_modules/(?!(react-native|styled-components|expo-asset|expo-constants|@unimodules|react-native-unimodules|expo-font|react-native-svg|@expo/vector-icons|styled-system|react-native-vector-icons|@react-native-aria/interactions|@react-native-aria/button|@react-native-aria/switch|@react-native-aria/toggle|@react-native-aria/utils|@react-native-aria/radio)/)"
```
theankurkedia, line 94: I think visually-hidden should also be added here.

src/components/primitives/Radio/Radio.tsx
```diff
@@ -23,85 +27,104 @@ const Radio = ({ icon, children, ...props }: IRadioProps, ref: any) => {
...
+
   // only calling below function when icon exist.
   const sizedIcon = () =>
     //@ts-ignore
     React.cloneElement(icon, {
       size,
       color:
         //@ts-ignore
-        icon.props.color ?? checked
-          ? isDisabled
```
theankurkedia, line 34: I am not sure if we are handling isDisabled inside useRadio? If not then we will have to manually convert it from isDisabled to disabled.
Because isDisabled is a prop, while disabled is not.

vidhi499: Yes, isDisabled is handled from useRadio hook. It sends disabled inside inputProps

package.json
vidhi499: Visually-hidden is directly imported from react-aria, so it will not create any issue.

PR #3486, opened 2021-02-17
fix: removed child component name check

MD-REHMAN:


MD-REHMAN:
Include changes on Component on Accordion, NumberInput, Tabs and PinInput

PR #3546, opened 2021-03-16
Fix/v3 minor fixes

MD-REHMAN:


MD-REHMAN:
Can cause a lot of conflicts. Moving required to separate PR.

PR #3561, opened 2021-03-23
fix: v3 Form control accessibility

intergalacticspacehighway:
This PR improves FormControl and Input accessibility. Tested on VoiceOver and added testcases for the same.

  - `nativeID` passed to the form control will be passed to the form input directly. 
  - `FormLabel` will have `htmlFor` that points to the `nativeID` of the form input.
  - `FormErrorMessage` adds `aria-describedby` and `aria-invalid` pointing to
    the form input.
  - `FormHelperText` adds `aria-describedby` pointing to
    the form input.
  - `FormHelperText` adds/extends `aria-describedby` pointing to the form input.

### DIFF
--- a/src/components/primitives/Input/index.tsx
+++ b/src/components/primitives/Input/index.tsx
@@ -27,8 +27,8 @@ import { useThemeProps, useToken } from '../../../hooks';
 import { themeTools } from '../../../theme';
 import { useHover } from '@react-native-aria/interactions';
 import {
-  FormControlContext,
-  IFormControlContext,
+  useFormControl,
+  useFormControlContext,
 } from '../../composites/FormControl';
 
 const StyledInput = styled(TextInput)<IInputProps>(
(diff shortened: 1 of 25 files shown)


src/components/primitives/Input/index.tsx
```diff
@@ -192,7 +197,7 @@ const Input = (
                 transform: [{ translateY: slideAnim, translateX: 4 }],
               }}
             >
-              <Flex {...newProps} bg="transparent">
+              <Flex {...newProps} nativeID={undefined} bg="transparent">
```
intergalacticspacehighway, line 200: @Ajamuar this is a bit hack which I have discussed with @MD-REHMAN , we'll need to refactor Input. 
We'll have to select which props should be passed into Flex and which into Input.

Ajamuar (approved):

PR #3567, opened 2021-03-23
Feat/v3 pre beta button

Ajamuar:
Changed the style of button to match that of Material Design.


MD-REHMAN:
@Ajamuar Shouldn't the branch would be `v3-pre-dev`

PR #3599, opened 2021-03-31
V3 select refactor 

intergalacticspacehighway:


src/components/primitives/Select/SelectItem.tsx
```diff
@@ -1,64 +1,48 @@
...
+      <Actionsheet.Item
+        {...props}
         onPress={() => {
           if (!isDisabled) {
-            onValueChange(value, currentIndex);
+            onValueChange(value);
             closeMenu && closeMenu();
           }
         }}
+        // endIcon={
```
intergalacticspacehighway, line 31: endIcon causes maximum callstack error so used HStack instead. Should be fixed after @MD-REHMAN's action sheet changes

Ajamuar (approved):

PR #3624, opened 2021-04-06
Add testID for Tabs

sepeterson:
Enables use of testID with tabs. This'll clear up some issues we had e2e testing screens w/ tabs in our app.


TwFlem:
Awesome! It looks like this will make the tabs consistent with the rest of of the native base components where they accept a testID. With this, users will be able to rely on testID instead of having to rely on xpaths for automated testing where the tabs are involved.

hamzaawais96:
Any update on this PR?

PR #3659, opened 2021-04-19
Fix/v3 pre beta list

Ajamuar:


rayan1810 (changes requested):

example/storybook/stories/components/primitives/Box/composition.tsx
```diff
@@ -23,6 +23,8 @@ export const Composition = () => {
                 'https://www.holidify.com/images/cmsuploads/compressed/Bangalore_citycover_20190613234056.jpg',
             }}
             alt="image"
+            w="100%"
+            h="100%"
```
MD-REHMAN, line 27: It's won't we required. Can you try this with latest pull?

PR #3688, opened 2021-05-12
Add `number` type to Spinner.size

davwheat:
Providing a number to `size` is supported. It can be used for specifying custom spinner sizes.


davwheat:
Not stale.

PR #3694, opened 2021-05-16
Update peer deps to make compatible with RN 0.63

davwheat:
Without this change, `npm audit --fix` will not work in projects that use NB v3 and RN 0.63.x, introducing possible security vulnerabilities.


davwheat:
Not stale.

PR #3707, opened 2021-05-24
v3 fix for Actionsheet Animated.event()

yduman:
Hi,

I discovered while using latest native-base v3 that when you trigger the animation for the Actionsheet component, the logs on Metro spam you with "Animated.event now requires a second argument for options". I believe that this PR should fix this. I couldn't test this though since the examples folder refused to work on my machine. I found this while using the Select component which makes use of the Actionsheet component.


intergalacticspacehighway:
@yduman Thanks for working on it!

Issue #3774, opened 2021-06-29
Box With Linear gradient example produces warning.


btrewern:
If you recreate the
**Box** -> **With Linear gradient** - https://docs.nativebase.io/box/#with-linear-gradient
with the current version of NativeBase (3.0.3) you get the following warning:

"Node of type rule not supported as an inline style"

This is on the
Android Emulator running Android 11
or Note 9 running Expo Go 2.19.6

rayan1810:
This issue should be resolved with next release, fix merged in PR - https://github.com/GeekyAnts/NativeBase/pull/3777

PR #3777, opened 2021-06-30
Fix: Box With Linear gradient example produces warning. 

cyrus25:
https://github.com/GeekyAnts/NativeBase/issues/3774

### DIFF
--- a/src/components/primitives/Box/index.tsx
+++ b/src/components/primitives/Box/index.tsx
@@ -52,7 +52,15 @@ const Box = ({ children, ...props }: IBoxProps, ref: any) => {
           y: lgrad.end[1],
         };
       }
-      if ('bg' in safeAreaProps) delete safeAreaProps['bg'];
+      const backgroundItems = [
+        'bg',
+        'bgColor',
+        'background',
+        'backgroundColor',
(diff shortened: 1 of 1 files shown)


src/components/primitives/Box/index.tsx
```diff
@@ -52,6 +52,7 @@ const Box = ({ children, ...props }: IBoxProps, ref: any) => {
           y: lgrad.end[1],
         };
       }
+      if ('bg' in safeAreaProps) delete safeAreaProps['bg'];
```
rayan1810, line 55: @cyrus25 we need to delete background, bgColor and backgroundColor as well. Please check line 23.

rayan1810 (changes requested):

src/components/primitives/Box/index.tsx
```diff
@@ -52,6 +52,15 @@ const Box = ({ children, ...props }: IBoxProps, ref: any) => {
           y: lgrad.end[1],
         };
       }
+      const backgroundItems = [
```
rayan1810, line 55: Can we follow proper naming , please use backgroundColorProps.

Issue #3811, opened 2021-07-08
Android: Custom fonts not rendered with correct font when using fontWeight/fontStyle


chaaya:
**Describe the bug**
Bold/SemiBold/Italic fonts not working on Android when you are using custom fonts.

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a custom font set to NativeBase
2. Render ```<Text fontWeight="bold">Hello world</Text>```
3. The Text gets rendered as the default device system font.
4. See screenshots

**Expected behaviour**
The <Text/> component should render the correct font/font-weight/font-style based on theme configuration.

**Screenshots**
***iOS (working)***
![image](https://user-images.githubusercontent.com/7448249/124925116-6984de00-dffc-11eb-804e-babafc12674c.png)
***Android (not working)***
![image](https://user-images.githubusercontent.com/7448249/124925152-71448280-dffc-11eb-8c50-9d5ad0e125f8.png)

**Platform**
 - Device: Samsung S10
 - OS: Android

**Additional information**
The correct font gets rendered  when I removed the fontWeight and fontStyle prop from <StyledText /> component in the source code. But that might break for users who doesnt have a custom font. My suggestion to the solution is to check if the user has not configured a custom font than add the fontWeight and fontStyle prop to the component internally otherwhise just omit them.

Would be happy to help on this issue and create PR but I am going on vacation now and wont be able to help you out right now.

intergalacticspacehighway:
@chaaya thanks for reporting, this has been fixed in the latest version.

ENT108:
It works only for iOS, Android doesn't respect `fontWeight` prop. Please re-open.

facuescobar:
Same here, please reopen it, font is not rendering correctly on Android

olivan94:
fontWeight prop still not working with custom fonts on Android

jimpala:
Still not working for me either, having to directly reference the name of the fontFamily at desired weight

nagydan1:
Still a problem, please reopen

epshtielsl:
@intergalacticspacehighway I am also seeing this issue in v3.4.20.  Font weights working in iOS but not in Android.

minhtamwebmaster:
Still a problem, please reopen

ccdwyer:
Still seeing this issue as well.

cjshearer:
[...]

PR #3821, opened 2021-07-12
fix: android custom fonts not working

intergalacticspacehighway:
Fixes - https://github.com/GeekyAnts/NativeBase/issues/3811

### Description
Android behaves slightly different when fontWeight and a custom font family is passed. Even though custom font family matches, it attaches system fonts when fontWeight is passed. To solve this, if a custom font family is resolved, we prevent passing fontStyle and fontWeight to the underlying Text component.

### Testing
Added testcases for the same.


package.json
```diff
@@ -71,7 +71,7 @@
     "postinstall-postinstall": "^2.1.0",
     "prettier": "^2.0.5",
     "react": "16.11.0",
-    "react-native": "0.62.2",
+    "react-native": "0.63.2",
```
Ajamuar, line 74: Are we bumping the version?

src/theme/base/typography.ts
```diff
@@ -66,6 +66,45 @@ const typography = {
   //       italic: 'Roboto-BoldItalic',
   //     },
   //   },
+
+  //   Montserrat: {
```
Ajamuar, line 70: Are these comments required?

package.json
intergalacticspacehighway: Yes, this version upgrade was due since we added Pressable component which is released in RN 0.63. Upgraded it in this PR to fix the failing of testcases.

src/theme/base/typography.ts
intergalacticspacehighway: Yes, actually we can remove it but kept it to make it easier to test custom fonts. 
We're loading Montserrat in example app. Just uncommenting will make it work here. Let me know if needs to be removed!

Ajamuar (approved):

PR #3841, opened 2021-07-15
Fix/breadcrumb

vidhi499:
## Summary
Earlier was unable to use IsCurrent props in case of nested children.(Fix for that)
Added maxItems props to make breadcrumb collapsible.
Refactoring of API (used HStack instead of flex for simpler code)

## Changelog
fixed an existing bug
Refactoring of API (used HStack instead of flex for simpler code)
Made breadcrumb collapsible


src/components/composites/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,29 +1,176 @@
...
+      onPress={() => {
+        props.setCollapsed(false);
+      }}
+    >
+      <ThreeDotsIcon size={4} />
+    </Pressable>
   );
 };
+
 export default React.memo(React.forwardRef(Breadcrumb));
```
MD-REHMAN, line 175: Remove `React.` for documentation.

MD-REHMAN (approved):

PR #3870, opened 2021-07-23
fix: custom toast with render

intergalacticspacehighway:
## Summary
This PR fixes passing id to a custom toast which uses render function in toast.show()

## Changelog

Fixes custom id not being passed in a custom toast

## Test Plan

Added a storybook example named CustomToastWithCustomId.


Ajamuar (approved):

zhangyu1818:
I found that in the document:
> Render a component toast component. Any component passed will receive 2 props: `id` and `onClose`.

but It seems there still doesn't pass 'onClose'.

PR #3896, opened 2021-07-28
fix: "out of place" props from Menu example

MattYoon:
## Summary

The `Select` component has `accessibilityLabel` and `placeholder` props directly copied and pasted from the [Select component example.](https://github.com/GeekyAnts/NativeBase/blob/master/example/storybook/stories/components/primitives/Select/Basic.tsx) The props are unnecessary since the example is to demonstrate the Menu component, not the Select component. Also the content of the props are "out of place", since the example is not about selecting a programming language.

## Changelog

[General] [Removed] - Remove `accessibilityLabel` and `placeholder` props from `Select` component in Menu Example

## Test Plan

It was directly tested from the snack player.

### DIFF
--- a/example/storybook/stories/components/composites/Menu/MenuPositions.tsx
+++ b/example/storybook/stories/components/composites/Menu/MenuPositions.tsx
@@ -32,6 +32,7 @@ export function Example() {
           bg: 'cyan.600',
           endIcon: <CheckIcon size={4} />,
         }}
+        accessibilityLabel="Select a position for Menu"
       >
         <Select.Item label="auto" value="auto" />
         <Select.Item label="Top Left" value="top left" />
--- a/example/storybook/stories/components/composites/Menu/MenuPositions.tsx
+++ b/example/storybook/stories/components/composites/Menu/MenuPositions.tsx
@@ -27,8 +27,6 @@ export function Example() {
       <Select
         selectedValue={position}
         mx={{ base: 0, md: 'auto' }}
-        accessibilityLabel="Select your favorite programming language"
-        placeholder="Select your favorite programming language"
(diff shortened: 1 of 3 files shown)


example/storybook/stories/components/composites/Menu/MenuPositions.tsx
```diff
@@ -27,8 +27,6 @@ export function Example() {
       <Select
         selectedValue={position}
         mx={{ base: 0, md: 'auto' }}
-        accessibilityLabel="Select your favorite programming language"
```
Ajamuar, line 30: Instead of removing this, can we add a message like `Select a position`?

MattYoon: I think it will be a bit odd if the `Menu` component itself doesn’t have a `accessibilityLabel` while the `Select` component has one. The example is for `Menu`, and `Select` is just there to help demonstrate `Menu` props.

rayan1810: But Select should also convey what it does or what it contains right?

MattYoon:
I changed `accessibilityLabel` based on the feedbacks, and I also fixed two more cases of `Select` copied and pasted.

example/storybook/stories/components/composites/Menu/MenuPositions.tsx
MattYoon: I think you are right. I fixed it, and also found two more cases, fixed them as well.

himanshu-satija (approved):

himanshu-satija (approved):

PR #3903, opened 2021-07-30
Fix/button api refactor

vidhi499:
- Added subtle variant
- Added _loading, _spinner, _disabled props
- Added spinnerPlacement (start/end) props in case of loading button
- Changed Icon button example
- Added aliases for start and end icon
- Pulled master branch

### DIFF
--- a/example/storybook/stories/components/primitives/Button/loading.tsx
+++ b/example/storybook/stories/components/primitives/Button/loading.tsx
@@ -8,8 +8,21 @@ export const Example = () => {
       space={2}
       alignItems={{ base: 'center', md: 'flex-start' }}
     >
-      <Button isLoading>Button</Button>
-      <Button isLoading isLoadingText="Submitting">
+      <Button
+        isLoading
+        spinnerPlacement="end"
+        _loading={{ bg: 'red.100', _text: { color: 'green.500' }, opacity: 1 }}
+        _spinner={{ color: 'blue.900', size: 'lg' }}
+      >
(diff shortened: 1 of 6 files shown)


example/storybook/stories/components/primitives/Button/loading.tsx
```diff
@@ -8,8 +8,21 @@ export const Example = () => {
       space={2}
       alignItems={{ base: 'center', md: 'flex-start' }}
     >
-      <Button isLoading>Button</Button>
-      <Button isLoading isLoadingText="Submitting">
+      <Button
```
himanshu-satija, line 11: @vidhi499 Let's also keep the basic examples with the more complex ones.

src/components/primitives/Button/Button.tsx
```diff
@@ -106,13 +123,26 @@ const Button = (
...
                 {isLoading && isLoadingText ? isLoadingText : children}
               </Box>
             ) : null}
+
             {endIcon && !isLoading ? endIcon : null}
+            {isLoading && spinnerPlacement === 'end' ? (
+              spinner ? (
+                spinner
+              ) : (
+                <Spinner
```
himanshu-satija, line 139: Let's follow DRY for this Spinner element. We can keep this in a variable and use the same for start and end both.

himanshu-satija (changes requested):

himanshu-satija (approved):

Issue #4126, opened 2021-09-27
Can't set height of Select component


xionxiao:
**Describe the bug**
I have a form with Input and Select. I want to make them same height. But when I set ```h='50px'```, Input component height changes but Select component remain unchanged.

<img width="301" alt="Screen Shot 2021-09-27 at 10 36 41 AM" src="https://user-images.githubusercontent.com/28080841/134838655-f89b3836-8bed-4f68-8977-98c9b9a5b533.png">

**Expected behaviour**

<img width="335" alt="Screen Shot 2021-09-27 at 10 34 24 AM" src="https://user-images.githubusercontent.com/28080841/134838729-47472c01-9eb7-4b10-ad1c-b02591fb08db.png">

**To Reproduce**
Here's the sample code
```typescript
export default function App() {
  return (
    <NativeBaseProvider>
      <VStack mt='16' space={5} alignItems="center">
        <Input h='50px' />
        <Select h='50px'>
          {['A', 'B', 'C'].map(i => <Select.Item label={i} value={i} />)}
        </Select>
      </VStack>
    </NativeBaseProvider>
  );
}
```

**Additional information**

As the inspect shows, the outer height changed to 50px, but the inner component.

<img width="441" alt="Screen Shot 2021-09-27 at 10 43 14 AM" src="https://user-images.githubusercontent.com/28080841/134839244-967a7758-cb20-40e3-a17a-80cbcc8ab26c.png">

**Platform**

- NativeBase version 3.2.0
- Device: iPhone, Android, Web

xionxiao:
Great! 
Very quick fix!

Issue #4122, opened 2021-09-25
Issue with the `Select` component.


obayit-ghaya:
The `Select` item seems to have an issue with a prop called `accessibilityDescribedBy`.

**To Reproduce**
Create an expo project, and install the dependencies, with:
```
expo init project-name
cd project-name
yarn add native-base styled-components styled-system
expo install react-native-svg
expo install react-native-safe-area-context
```

Then add a select component, my exact `App.js` is:
```
import React from 'react';
import { StatusBar } from 'expo-status-bar';
import { StyleSheet, Text, View } from 'react-native';

import {
  NativeBaseProvider, Box,
  Select,
  VStack,
  CheckIcon,
  Center,
} from "native-base";

export const Example = () => {
  let [service, setService] = React.useState("")
  return (
    <VStack alignItems="center" space={4}>
      <Select
        selectedValue={service}
        minWidth="200"
        accessibilityLabel="Choose Service"
        placeholder="Choose Service"
        _selectedItem={{
          bg: "teal.600",
          endIcon: <CheckIcon size="5" />,
        }}
        mt={1}
        onValueChange={(itemValue) => setService(itemValue)}
      >
        <Select.Item label="UX Research" value="ux" />
        <Select.Item label="Web Development" value="web" />
        <Select.Item label="Cross Platform Development" value="cross" />
        <Select.Item label="UI Designing" value="ui" />
        <Select.Item label="Backend Development" value="backend" />
      </Select>
    </VStack>
  )
}

export default function App() {
  return (
    <NativeBaseProvider>
      <View style={styles.container}>
        <Example />
        <StatusBar style="auto" />
      </View>
    </NativeBaseProvider>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
});
```
Run the app with `yarn web`, and the error should be in console.

**CodeSandBox/Snack link**
I couldn't replicate this in `snack.expo.dev`,  [here](https://snack.expo.dev/@obay.ghaya/nativebase-template). But the issue shows in my development machine :(

**Screenshots**
Start of the console:
![Screenshot from 2021-09-25 10-16-11](https://user-images.githubusercontent.com/90961606/134764594-7c11eb8a-25a0-43d6-8055-f9bcdf7acf66.png)
End of the console:
[...]

PR #4129, opened 2021-09-27
fix: height prop not working in select

intergalacticspacehighway:
Fixes https://github.com/GeekyAnts/NativeBase/issues/4126 and https://github.com/GeekyAnts/NativeBase/issues/4122


obayit:
Just curious, how can I know that this fix is now deployed to the `native-base` package on `npm`?

PR #4222, opened 2021-10-20
[FIX] Invalid prop flex of type string supplied to StyleSheet box, ex…

jgtoriginal:
## Summary

Error on `➜  NativeBase git:(master) vim src/components/primitives/Box/index.tsx`
Due to flex being passed wrong data. It makes Box brake on iPad.

image

## Changelog

Box now receives correct parameter when called on iPad.

[CATEGORY] [TYPE] - Message

## Test Plan

* Use Box on iPad, it brakes, use this branch it fixes.
* console.log `safeAreaProps` on src/components/primitives/Box/index.tsx:93 to see wrong argument being passed

### DIFF
--- a/src/components/primitives/Box/index.tsx
+++ b/src/components/primitives/Box/index.tsx
@@ -89,6 +89,9 @@ const Box = ({ children, ...props }: IBoxProps, ref: any) => {
       );
     }
   }
+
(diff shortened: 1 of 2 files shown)


jgtoriginal:
@surajahmed is this fixed? Any references you may share?

PR #4311, opened 2021-11-17
fix: export IModalProps from components Modal

Youngtao1999:
fix: export IModalProps from components Modal


Ajamuar (approved):

Ajamuar:
IModalProps has already been added in the `minor` branch. We will be releasing it in the next alpha release.

Youngtao1999:
My regret.

Issue #4176, opened 2021-10-11
Poor performance for the `Select` component


aimensasi:
**Describe the bug**
Poor performance for the `Select` component. When clicking on the select component, there is a very noticeable delay. The issue occurs again when selecting an item from the action sheet displayed. Note that, The issue appears on Android devices more clearly than on IOS.

**Sample Code**

```
<Select
    selectedValue={values.size}
    minWidth="200"
    accessibilityLabel="Choose Company Size"
    placeholder="Choose Company Size"
    mt={1}
    onValueChange={(itemValue) => setFieldValue("size", itemValue)}
  >
    <Select.Item label="1 to 9 employees" value="1 to 9 employees" />
    <Select.Item label="10 - 49 employees" value="10 - 49 employees" />
    <Select.Item label="50 - 249 employees" value="50 - 249 employees" />
    <Select.Item label="250 or more employees" value="250 or more employees" />
  </Select>
```

**Screenshots**
https://user-images.githubusercontent.com/13252355/136733406-c38e3c58-3e5c-48c2-9544-ec0663295006.mov

**Platform**

- NativeBase version: 3.20
- Device: Samsung A12
- OS: Android 11
- React Native version 0.65.1
- Expo SDK version 42.0.1
- node 14.15.4
- yarn: 1.22.15

**Additional information**

The issue seems to get worst when using Formik or react-form-hook, but it is still there even without using them.
The number of items in the list is not the cause of the issue, as I have tried from 2 to 14 items and the issue is still there.

bencreynolds:
Same issue here, but only when rendering dozens of items. Rendering like 5 items (using map) does not cause an issue.

MD-REHMAN:
We have done some enhancements to fix `Select`'s performance. Do let us know if you're still facing issues.

ksi9302:
Hi, I'm using NativeBase 3.3.4 I'm still facing big performance issue with select component.
When I select the item on the action sheet, it seems like it waits till all the triggered actions are complete and the relevant components are ready to be rendered.
In other words, it feels asynchronous.

Viraj-10:
Hi @ksi9302, Can you share small code and environment details.

Viraj-10:
Hi @ksi9302, Any Updates?

danielantelo:
On latest NativeBase having a select such as:

```javascript
         <Select
            selectedValue={gender || ''}
            placeholder={'select...'}
            minWidth={220}
[...]

PR #4327, opened 2021-11-22
feat: update select to fix performance issues

MD-REHMAN:
## Summary 

Update Select to optimize Select's performance #4176.


bencreynolds:
Probably a noobie question but when would a change like this be merged? Without it, I can't use the library.

Issue #4168, opened 2021-10-07
FlatList does not support _contentContainerStyle


trajano:
I tried to add

```
      _contentContainerStyle={{
        p: "4",
      }}
```
Works well on ScrollView but does not appear to be recognized (nor is it in the types) for `FlatList`

For now I worked around this gap (and the fact that IFlatListProps is not exported and that FlatList does not do the generics as in RN) using this

```
import {
  FlatList as NBFlatList,
  IScrollViewProps,
  useStyledSystemPropsResolver,
} from "native-base";
import type { FlatListProps as RNFlatListProps } from "react-native";
import React from "react";
export type FlatListProps<ItemT> = Omit<
  Parameters<typeof NBFlatList>[0],
  "data" | "renderItem"
> &
  RNFlatListProps<ItemT> &
  IScrollViewProps["_contentContainerStyle"];
export function FlatList<ItemT = any>({
  _contentContainerStyle,
  contentContainerStyle,
  ...props
}: FlatListProps<ItemT>): JSX.Element {
  const resolved_ContentContainerStyle = useStyledSystemPropsResolver(
    _contentContainerStyle || {}
  );
  return (
    <NBFlatList
      {...props}
      contentContainerStyle={
        contentContainerStyle || resolved_ContentContainerStyle
      }
    />
  );
}
```

surajahmed:
@trajano This issue has been resolved in version `3.3.0` or later. Thanks for reporting!

PR #4337, opened 2021-11-22
feat: added _contentContainerStyle for flatList

MD-REHMAN:
## Summary
added _contentContainerStyle for flatList [#4168]


thispastwinter:
May I suggest the following for the _contentContainerStyle type? 

`Omit<IFlatListProps, "data", "renderItem">`

Currently, I'd love to be able to use this to get the added benefits of the utility props, however it expects me to pass data and renderItem.

PR #4350, opened 2021-11-25
Added disableDynamicWidth option to avoid screen orientation from propagating many state changes.

richardwu:
## Summary

Screen orientation that changes the window's width will propagate breakpoint calculation + theme sizing.

This is expected: however, this can cause very laggy behavior if dynamic breakpoints/sizing is not necessary.

## Changelog

[Fixes][Performance] - Added `disableDynamicWidth` theme option to disable dynamic breakpoint calculation that can cause performance issues with screen orientation

## Test Plan

I've tested this on my local device and DOM updates are much faster when switching between landscape/portrait.

### DIFF
--- a/src/core/NativeBaseProvider.tsx
+++ b/src/core/NativeBaseProvider.tsx
@@ -16,7 +16,7 @@ import {
   NativeBaseConfigProvider,
 } from './NativeBaseContext';
 import { useToast } from '../components/composites/Toast';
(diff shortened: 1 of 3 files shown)


Ajamuar:
@richardwu We are currently looking at some options to reach a similar objective. It might take a little bit of time but we are considering your solution for the same. We request your patience on this issue. We will get back to you ASAP.

PR #4377, opened 2021-12-06
fix: switch toggling issue with value

MD-REHMAN:


codingslash76:
`onToggle` return `undefined`

PR #4383, opened 2021-12-07
fix: typing color by custom themes

movibe:
## Summary
Typing adjustment for custom themes

## Changelog

[General] [Fixed] - Typing color by custom themes

## Test Plan

When creating a custom theme, the color token was not received by auto complete


movibe:
Thanks!!!

PR #4454, opened 2021-12-24
fix: info toast long title

Viraj-10:
## Summary
Info Icon and close button was going out of the screen for android and ios device when title of the info toast is long.

### DIFF
--- a/src/components/composites/Toast/Toast.tsx
+++ b/src/components/composites/Toast/Toast.tsx
@@ -271,15 +271,15 @@ export const ToastProvider = ({ children }: { children: any }) => {
           <Alert
             maxWidth="100%"
             alignSelf="center"
+            flexDirection="row"
             status={status ?? 'info'}
             variant={variant as any}
             accessibilityLiveRegion={accessibilityLiveRegion}
             {...rest}
           >
-            <VStack space={1} flexShrink={1} w="100%">
+            <VStack space={1} flexShrink={1}>
               <HStack
                 flexShrink={1}
-                space={2}
                 alignItems="center"
                 justifyContent="space-between"
               >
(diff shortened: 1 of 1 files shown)


src/components/composites/Toast/Toast.tsx
```diff
@@ -271,15 +271,15 @@ export const ToastProvider = ({ children }: { children: any }) => {
           <Alert
             maxWidth="100%"
             alignSelf="center"
+            flexDirection="row"
             status={status ?? 'info'}
             variant={variant as any}
             accessibilityLiveRegion={accessibilityLiveRegion}
             {...rest}
           >
-            <VStack space={1} flexShrink={1} w="100%">
```
rayan1810, line 279: No need to remove w="100%"

src/components/composites/Toast/Toast.tsx
```diff
@@ -271,15 +271,15 @@ export const ToastProvider = ({ children }: { children: any }) => {
...
+            <VStack space={1} flexShrink={1}>
               <HStack
                 flexShrink={1}
-                space={2}
```
rayan1810, line 282: Why did we removed the spacing here?

rayan1810 (changes requested):

src/components/composites/Toast/Toast.tsx
Viraj-10: it was creating a issue with info icon and close button. Both were not having same distance from text when title is long on IOS and Android. small UI change suggested by Himanshu.

src/components/composites/Toast/Toast.tsx
Viraj-10: added w="100%"

rayan1810 (approved):

PR #4472, opened 2021-12-29
improvement: allow as prop of Icon to be themed

johnf:
## Summary

```as``` is pulled out in the parameters to Icon so it cannot be themed. For my particular example I want to be able to do something like
```js
const customTheme = extendTheme({
  components: {
    Icon: {
      defaultProps: {
        as: FontAwesome5,                                                                                                                       
      },                                                                                                                                        
    },                                                                                                                                          
  },
})             
```
## Changelog

General Added - Make **as** themeable for **Icon**

## Test Plan

This is a fairly minor change and I've tested locally in my own app


surajahmed:
@johnf Thanks for working on this ❤️ .
We'll test and merge it in the coming release.

johnf:
@surajahmed bumping since it's gone stale

Issue #4469, opened 2021-12-28
FlatList not properly typed


MoSattler:
`FlatList`'s `renderItem` does not infer the type from the `data` prop. On the other hand, the regular react-native's flatlets does.

```typescript
<FlatList
  data={["abc"]}
  renderItem={({ item })= > item} // React-native's inferred `item` to be `string`. In native-base it's `any`
/>
```

Same goes for other props like `keyExtractor`.

Viraj-10:
Hi @MoSattler,
Thanks for reporting the issue.
It is fixed now and PR is created. #4478 
It will be included in upcoming release. 😄

Viraj-10:
This issue is fixed in Version 3.3.0 of NativeBase

PR #4478, opened 2021-12-30
Fix: flatlist prop type

Viraj-10:
## Summary
github Issue:- [https://github.com/GeekyAnts/NativeBase/issues/4469](url)
NativeBase FlatList was not properly typed when it's compared with react-native FlatList.

### DIFF
--- a/src/components/basic/FlatList/FlatList.tsx
+++ b/src/components/basic/FlatList/FlatList.tsx
@@ -10,26 +10,28 @@ import { useHasResponsiveProps } from '../../../hooks/useHasResponsiveProps';
 
 const StyledFlatList: any = makeStyledComponent(RNFlatList);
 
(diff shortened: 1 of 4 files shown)


src/components/basic/FlatList/FlatList.tsx
```diff
@@ -10,7 +10,10 @@ import { useHasResponsiveProps } from '../../../hooks/useHasResponsiveProps';
 
 const StyledFlatList: any = makeStyledComponent(RNFlatList);
 
-export const FlatList = forwardRef((props: IFlatListProps, ref: any) => {
+const FlatListComponent = <ItemT extends any>(
```
MoSattler, line 13: is the `extends any` necessary? 
```suggestion
const FlatListComponent = <ItemT>(
```

PR #4539, opened 2022-01-16
fix: infinite loop in Image component

kaykdm:
## Summary
Fix infinite loop in Image component
[This fix](https://github.com/GeekyAnts/NativeBase/pull/4529) is causing infinite loop in Image component when source is changed.
You can see the error in the example of `Image/FallbackSupport` in storybook.
I found this issue when I was using Modal to display Image in react native by passing source via props when it is opened.

this is the error
```
Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

## Changelog

[CATEGORY] [TYPE] - Message

### DIFF
--- a/src/components/primitives/Image/index.tsx
+++ b/src/components/primitives/Image/index.tsx
@@ -28,7 +28,8 @@ const Image = (props: IImageProps, ref: any) => {
       finalSource.current = { uri: src };
     }
     return finalSource.current;
-  }, [source, src]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [source?.uri, src]);
 
   const [renderedSource, setSource] = useState(getSource());
   const [alternate, setAlternate] = useState(false);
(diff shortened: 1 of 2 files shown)


src/components/primitives/Image/index.tsx
```diff
@@ -28,7 +28,8 @@ const Image = (props: IImageProps, ref: any) => {
       finalSource.current = { uri: src };
     }
     return finalSource.current;
-  }, [source, src]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kaykdm, line 31: Since setting `source` in useEffects' deps is causing an error, I am using `source.uri` and disabling `react-hooks/exhaustive-deps` for `source`

surajahmed:
@kaykdm Thank you for working on this ❤️

PR #4567, opened 2022-01-22
fix(Image): fix maximum update depth exceeded error

shanedaugherty:
## Summary

This PR fixes an issue where the `useEffect` hook in the image component runs recursively until React throws a `Maximum update depth exceeded error`.

 It solves the problem by refactoring the various hooks involved with determining, keeping in sync, and storing the image source. The most notable change is that `setSource` is only called when the source actually changes.

Note: JSON.stringify is used to compare the previous / next value because other methods of comparison are tricky due to the state being an object that many people will not remember to memo. 

## Changelog
PATCH FIX - fixed bug in the Image component causing some apps to run an infinite loop

## Test Plan
I honestly do not know how to test this one - perhaps there is a way to write a jest test that catches the `Maximum update depth exceeded error`?

### DIFF
--- a/src/components/primitives/Image/index.tsx
+++ b/src/components/primitives/Image/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, memo, forwardRef, useCallback, useRef } from 'react';
+import React, {useState, memo, forwardRef, useCallback, useRef, useMemo, useEffect} from 'react';
 import { Image as RNImage } from 'react-native';
 import Text from '../Text';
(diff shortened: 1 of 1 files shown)


kaykdm:
@shanedaugherty 
Can you still reproduce this error in latest release (3.3.4) ?

surajahmed:
@shanedaugherty This issue has been resolved with this PR (https://github.com/GeekyAnts/NativeBase/pull/4539). Thank you for working on this.

Issue #4587, opened 2022-01-29
To-Do Sample is Broken


rlingineni:
### Description

To-Do Sample broken, checkboxes resetting

### CodeSandBox/Snack link

https://docs.nativebase.io/todo-list

### Steps to reproduce

1. Go to [Docs Site](https://docs.nativebase.io/todo-list) 
2. Click the to-do list, and check a few things
3. Clicking one checkbox will uncheck others

### NativeBase Version

3.3.1

### Platform

- [ ] Android
- [ ] CRA
- [X] Expo
- [X] iOS
- [ ] Next

### Other Platform

Web

### Additional Information

After some investigation, I found that in the `handleStatusChange` function, the `list` state variable is getting reset to the initial state.

It seems that the `onChange` handler for the checkbox isn't somehow getting binded to latest version of the `list` state. I tested this by adding a Button and the `onPress` handler, and the behavior worked as expected.

```jsx
 <HStack
  w="100%"
   justifyContent="space-between"
   alignItems="center"
   key={itemI}
 >
                <Button
                  onPress={() => {
                  // calling this in the OnPress Handler works as expected, but seems like onChange handler in the Checkbox doesn't work
                  handleStatusChange(itemI)
                  }
                  bg={"white"}
                  color={item.isCompleted ? "black" : "red"}
                >
                  {item.title}
                </Button>
                <Checkbox
                  isChecked={item.isCompleted}
                  onChange={(isChecked) => handleStatusChange(itemI)}
                  value={item.title}
                ></Checkbox>
</HStack>
```

Can you please confirm the Checkbox handler is working as expected?

surajahmed:
@rlingineni Thanks for reporting! The To-Do sample is fixed in version `3.3.5`.

PR #4588, opened 2022-01-29
fix(example): fix todo hooks state

yuxizhe:
## Summary
https://docs.nativebase.io/todo-list

There is a bug on the website. 

Jan-29-2022 16-17-58

fix this hooks state problem.

fix https://github.com/GeekyAnts/NativeBase/issues/4587

## Changelog

fix(example): fix todo hooks state

### DIFF
--- a/example/storybook/stories/examples/TodoApp.tsx
+++ b/example/storybook/stories/examples/TodoApp.tsx
@@ -36,8 +36,8 @@ export const Example = () => {
   const [inputValue, setInputValue] = React.useState('');
 
   const addItem = (title: string) => {
-    setList([
-      ...list,
+    setList((prevList) => [
+      ...prevList,
       {
         title: title,
         isCompleted: false,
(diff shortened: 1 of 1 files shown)


rlingineni:
@yuxizhe  - this seems a little hacky.

I don't think we should be using the callback in the sample to the set the state. I've been looking at more to-do samples online, and this PR doesn't seem like the natural way to do things.

I think there might be bug in the onPress handler for the `Checkbox`.

[Dev To Blog Post](https://dev.to/joelynn/how-to-build-a-react-crud-todo-app-edit-todo-46g6)

[Simple To-Do](https://towardsdatascience.com/build-a-simple-todo-app-using-react-a492adc9c8a4)

```
setList((prevList) => {})
```

For example, I used a button instead of the checkbox with the existing code, and it seemed to work as expected, but calling from the `onChange` handler of the Checkbox seems to fail. 

Take a look at this issue: #4587

PR #4592, opened 2022-01-30
Exposing underlying Image properties on Avatar via _image

kav:
## Summary
I have several use cases where I need to modify properties on the Avatars underlying image. This change adds _image to the avatar in a similar fashion to other cases e.g. _text
## Changelog

[General] [Added] - Added `_image` property to Avatar to expose the underlying image properties

## Test Plan
I've been using the patch successfully to drive underlying image tinting based on a focus property. I'm happy to add automated tests if someone can point me in the right direction. 

One area I'd question, and ask reviewsers to chime in on, is should the _image props allow override of all props provided by the avatar or are there some that should never be overiden? I.e. is `{..._image}
` placed correctly in the props order on the Image tag.

### DIFF
--- a/src/components/composites/Avatar/Avatar.tsx
+++ b/src/components/composites/Avatar/Avatar.tsx
@@ -1,12 +1,12 @@
-import React, { memo, forwardRef } from 'react';
-import { Box, Image, Text } from '../../primitives';
-import { usePropsResolution } from '../../../hooks/useThemeProps';
-import type { IAvatarProps } from './types';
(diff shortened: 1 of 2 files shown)


kav:
@intergalacticspacehighway @surajahmed or any other maintainer; any chance I could get a review or some notes on this one? Happy to open an issues as well if needed. I've got a few more coming and I'd love to sequence them if possible

surajahmed:
@kav Thanks for working on this ❤️ ! 
We've already made these changes in version `3.4.0`.

PR #4597, opened 2022-02-01
fix: todo example

Viraj-10:
## Summary

Fixing Todo Example.


rlingineni:
@surajahmed - can you take a look at the comments on this PR, this code looks to be the same as the one in this PR: https://github.com/GeekyAnts/NativeBase/pull/4588#issuecomment-1024963299

Do you think using the callback is the proper way, because it seems like a workaround? It seems that there's something wrong with the checkbox handler and binding, and if you could verify, that would be awesome.

PR #4620, opened 2022-02-09
fix: sectionlist type issue

Viraj-10:
### Fixes

- SectionList was not properly typed.
- fixed using assertion same as flatList.


src/components/basic/SectionList/types.ts
```diff
@@ -1,9 +1,11 @@
 import type { StyledProps } from '../../../theme/types';
 import type { SectionListProps } from 'react-native';
 import type { PlatformProps } from '../../types';
-
+type DefaultSectionT = {
+  [key: string]: any;
+};
 // TODO: any need to fixed
```
JaapWeijland, line 7: This todo may be removed right?

Issue #4617, opened 2022-02-09
set useRNModalOnAndroid from ActionSheet _backdrop props


s0rta:
### Description

Action sheets on android right now do not allow for interaction for elements behind the action sheet even when the overlay is disabled, I tracked this issue to the prop useRNModalOnAndroid, which when going from a modal to an overlay, is impossible to set by hand. We are building an app that requires an action sheet and not allowing the user to interact with the elements above the action sheet at the same time would be an antipattern. Would it be possible to allow for the setting of useRNModalOnAndroid from the _backdrop ActionSheet props? 

### Problem Statement

An action sheet on android with the overlay disabled still does not allow for user interaction.

### Proposed Solution or API

Allowing for the useRNModalOnAndroid prop to be set from the Action Sheet backdrop props

`_backdrop{{useRNModalOnAndroid: false}}`

### Alternatives

I've got it jerry rigged in a local build but imagine other folks would like to be able to use an action sheet as well as the rest of the app at the same time.

s0rta:
Another possible solution would be to pass overlayVisible down to the overlay as opposed to just the backdrop and use that to decide if a RN modal should be used or not.

surajahmed:
@s0rta Thanks for pointing this out. We'll look into this.

DavidAmyot:
@surajahmed Had the same issue on android, for now just setting `useRNModalOnAndroid = true` in the Overlay.tsx component fixed the issue for me.

surajahmed:
@DavidAmyot We'll make this change. Thanks a lot!

s0rta:
@DavidAmyot setting it to true allowed for interaction of elements behind the action sheet? It seems to me that the android specific modal was causing the problems. Also, after 3.4.0, disabling the android overlay option does not work any more. was there a fundamental change to how overlays work?

PR #4656, opened 2022-02-18
Android modal overlayVisible interactivity 

s0rta:
## Summary
allowing overlayVisible to reach the actual overlay, thus allowing for interaction outside of a modal while overlayVisible is set to false, this was already possible on iOS, but on android, because of using a react native modal, even with the overlay not visible, only the modal was interactive. 

## Changelog

[Android] [Changed] - An app with an open modal but with overlayVisble set to false is fully interactive, inside and outside of the modal.

## Test Plan
Create a screen with a modal (easy to test with an action sheet) and a scroll view. When overlayVisible is set to true, the overlay should show and the scroll view should not be interactive, when set to false, the overlay should not be visible and the scroll should be interactive.

### DIFF
--- a/src/components/composites/Modal/Modal.tsx
+++ b/src/components/composites/Modal/Modal.tsx
@@ -79,6 +79,7 @@ const Modal = (
       onRequestClose={handleClose}
       isKeyboardDismissable={isKeyboardDismissable}
       animationPreset={animationPreset}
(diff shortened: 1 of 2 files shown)


s0rta:
#4617 is the issue trying to be solved.

PR #4694, opened 2022-03-04
Allow for inputProps.style to take effect

droplet-tech:
## Summary

`inputProps.style` is not passed into input but is a valid prop 

```js
    if (propStyle) {

// Gets placed here 
// useStyledSystemPropsResolver.ts
      return { style: [styleSheet.box, propStyle], dataSet };
    } else {
      return { style: styleSheet.box, dataSet };
    }

```

This will allow better flexibility for users to allow the parent component to have final say in the style when needed. 

For example: `{ "resize": "auto" }` for TextArea in web


joseito-terence:
Is there a way to set { "resize": "auto" } for web on Textareas?

PR #4712, opened 2022-03-09
Decoupling Input component 

amars29:


src/components/primitives/Input/InputBase.tsx
```diff
@@ -138,7 +139,7 @@ const InputBase = (
           }
         : {})}
       ref={mergeRefs([ref, _ref, wrapperRef])}
-      style={Platform.OS === 'web' ? { outline: 'none' } : {}}
+      style={Platform.OS === 'web' ? webInputBase : {}}
```
rayan1810, line 142: @amars29 can't we implement this in some other way instead of creating a new prop?

amars29: @rayan1810 I think we can use it here for now. I will replace it if I find a better way.

PR #4713, opened 2022-03-09
fix: decoupling and added pseudo prop in select

mayank-96:
# Select

- Theme decoupling
- New pseudo prop  - `_actionSheetBody`
  ( props to be passed to underlying Flatlist in ActionSheet.Content. )

### DIFF
--- a/src/components/primitives/Select/Select.tsx
+++ b/src/components/primitives/Select/Select.tsx
@@ -69,7 +69,7 @@ const Select = (
     customDropdownIconProps,
     _actionSheetContent,
     _actionSheetBody,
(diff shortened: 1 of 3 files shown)


src/components/primitives/Select/Select.tsx
```diff
@@ -79,6 +68,8 @@ const Select = (
     optimized,
     customDropdownIconProps,
     _actionSheetContent,
+    _actionSheetBody,
+    unstyledSelecWebtStyles,
```
rayan1810, line 72: replace unstyledSelecWebtStyles with _webSelect

PR #4715, opened 2022-03-09
Decoupling theme in Slider component

amars29:


src/components/primitives/Slider/SliderTrack.tsx
```diff
@@ -48,10 +48,10 @@ const SliderTrack = ({ children, ...props }: ISliderTrackProps, ref?: any) => {
...
       {...trackProps}
       {...trackStyle}
-      paddingY={!isVertical ? '12px' : undefined}
-      paddingX={isVertical ? '12px' : undefined}
-      justifyContent="center"
-      alignItems="center"
+      paddingY={!isVertical ? '3' : undefined}
+      paddingX={isVertical ? '3' : undefined}
+      alignItems={_wrap.alignItems}
+      justifyContent={_wrap.justifyContent}
```
rayan1810, line 54: _wrap should be spread and all these should go inside that.

rayan1810: Also name it _pressable instead of _wrap

rayan1810: Also name it _pressable instead of _wrap

src/components/primitives/Slider/Slider.tsx
```diff
@@ -97,13 +97,7 @@ function Slider({ isDisabled, isReadOnly, ...props }: ISliderProps, ref?: any) {
...
   return (
     <SliderContext.Provider value={contextValue}>
-      <Box
-        {...wrapperStyle}
-        justifyContent="center"
-        ref={ref}
-        alignItems="center"
-        {...resolvedProps}
-      >
+      <Box {...wrapperStyle} ref={ref} {...resolvedProps}>
```
rayan1810, line 100: wrapperStyle should also be moved to theme

rayan1810, line 100: wrapperStyle should also be moved to theme

rayan1810 (changes requested):

PR #4773, opened 2022-03-23
Remove hardcoded peer dependency versions. Fixes #4575

### LINKED ISSUE #4575: Some issues with peers
### Description

peers  "react-native-safe-area-context", "react-native-svg" should not be version locked.  The @types should not be peer dependencies as they are not needed at runtime.

### CodeSandBox/Snack link

na

### Steps to reproduce

```
  "peerDependencies": {
    "@types/react": "*", // should not be here these are only used for development
    "@types/react-native": "*", // should not be here these are only used for development
    "react": "*",
    "react-dom": "*",
    "react-native": "*",
    "react-native-safe-area-context": "3.1.9", // should not be version locked or at least not locked to the point that 3.2 which is out cannot be used unless there's a real restriction.
    "react-native-svg": "12.1.0" //  // should not be version locked in case they release a new version and then nativebase becomes another limitting factor
  },
```

Please test with `npm` for the examples you're likely testing only on Yarn which means you will have less strict dependency checks unlike `npm`

### NativeBase Version

3.3.4

### Platform

- [ ] Android
- [ ] CRA
- [ ] Expo
- [ ] iOS
- [ ] Next

### Other Platform

_No response_

### Additional Information

_No response_

chrisbianca:
## Summary

I was unable to install NativeBase into a vanilla react-native project due to hardcoded dependency versions of react-native-safe-area-context and react-native-svg. This had previously been raised here: https://github.com/GeekyAnts/NativeBase/issues/4575

## Changelog

[General] [Fixed] - Removed hardcoded peer dependency versions

## Test Plan

I have tested in my own app and the NativeBase components work as expected.


trajano (changes requested):
I would recommend that instead of making it "*" at least specify a minimum viable version.

chrisbianca:
Looks like this has been addressed on master now

PR #4785, opened 2022-03-28
fix: space between Select dropdown and right border

amars29:


src/components/primitives/Select/Select.tsx
```diff
@@ -128,7 +128,7 @@ const Select = (
     ) : dropdownIcon ? (
       dropdownIcon
     ) : (
-      <ChevronDownIcon {...customDropdownIconProps} />
+      <ChevronDownIcon {...customDropdownIconProps} mr="3" />
```
ankit-tailor, line 131: Can we add `mr="3"` in `select.ts`?

ankit-tailor (approved):

PR #4792, opened 2022-03-28
fix: slider disabled state opacity issue

Viraj-10:
### Fix

- added `opacity` for the `disabled` state.

### DIFF
--- a/src/theme/components/slider.ts
+++ b/src/theme/components/slider.ts
@@ -21,9 +21,6 @@ export const SliderTrack = {
       _dark: {
         bg: 'muted.700',
       },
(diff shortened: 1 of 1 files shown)


mayank-96:
Screenshot 2022-03-29 at 11 36 25 AM
Screenshot 2022-03-29 at 11 37 41 AM
Slider Thumb is not looking similar to theme

mayank-96:
Also need to add `cursor: not-allowed` in disabled state for Slider Track and Slider Thumb

mayank-96 (approved):

ankit-tailor (approved):

PR #4840, opened 2022-03-30
feat: avoid spaces on Icon only button

grean:
## Summary

Since V3.3.8 a space has been added to the right of an icon (if it's icon start position). If there is no text in the button, the icon `appear` not correctly centered.
The fix test if children is `undefined`. If so it return null so there is no more spaces added.


grean:
Hi @MD-REHMAN,

When do you think this PR will be reviewed?

Thanks to you :)

PR #4921, opened 2022-04-12
added new fab props

shivamkawde18:
The fab component only accepts the following values for its placement prop:
"top-right" | "top-left" | "bottom-right" | "bottom-left".
Added new props for fab component-
->”top-center”
->”bottom-center”
->”center”
->”left-center”
->”right-center”

### DIFF
--- a/src/components/composites/Fab/types.tsx
+++ b/src/components/composites/Fab/types.tsx
@@ -6,7 +6,8 @@ export interface InterfaceFabProps extends InterfaceButtonProps {
    * Placement of the Fab
    * @default bottom-right
    */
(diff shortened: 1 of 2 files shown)


jd1378:
any news on this ?

jpike88:
I've noticed an issue: left '50%' does not perfectly centre the FAB.
Screen Shot 2022-07-12 at 09 46 59
.

jd1378:
> I've noticed an issue: left '50%' does not perfectly centre the FAB. Screen Shot 2022-07-12 at 09 46 59 .

It certainly will not, from my experience on web, you need to transformX it to the left by 50 percent so it perfectly center it

jd1378:
im not sure, but is it possible to wrap it in another component and margin from both sides ?

PR #5037, opened 2022-05-25
Update types.ts for primitive Input

fnarbona:
## Summary

This adds 'RefCallback' to ref property for primitive input in order to accommodate for libraries like react-hooks-form which pass a RefCallback type which is not currently accepted.


Viraj-10:
Hi @fnarbona, Thanks for your contribution.

Issue #5021, opened 2022-05-18
Option to Disable useRNModalOnAndroid on Actionsheet


markfuechec:
### Description

I'm building an actionsheet with native-base that is never closed. It sits on a screen with a map and a few buttons to interact with the map and is either in a barely-visible position or full size, toggled by user gestures. Unfortunately, this makes interacting with the map and buttons impossible on Android, unless I go into the native-base files and set useRNModalOnAndroid to false in the Modal component.

It would be cool if the actionsheet component had a property to set useRNModalOnAndroid to false.

### Problem Statement

Actionsheets use RN modals on Android. This causes the background to be non-interactive. I'd like the option to set that property to false.

### Proposed Solution or API

import React, { memo, forwardRef } from 'react';
import { Modal } from '../../composites/Modal';
import type { IActionsheetProps } from './types';
import { usePropsResolution } from '../../../hooks';
import { ActionSheetContext } from './ActionSheetContext';
import { useHasResponsiveProps } from '../../../hooks/useHasResponsiveProps';

const Actionsheet = (
  { children, hideDragIndicator = false, ...props }: IActionsheetProps,
  ref: any
) => {
  const {
    isOpen,
    disableOverlay,
    onClose,
    useRNModalOnAndroid, // Add this
    ...resolvedProps
  } = usePropsResolution('Actionsheet', props);

  //TODO: refactor for responsive prop
  if (useHasResponsiveProps(props)) {
    return null;
  }
  return (
    <Modal
      isOpen={isOpen}
      onClose={onClose}
      {...resolvedProps}
      overlayVisible={disableOverlay ? false : true}
      closeOnOverlayClick={disableOverlay ? false : true}
      ref={ref}
      useRNModalOnAndroid={useRNModalOnAndroid} // Add this
    >
      <ActionSheetContext.Provider value={{ hideDragIndicator }}>
        {children}
      </ActionSheetContext.Provider>
    </Modal>
  );
};

export default memo(forwardRef(Actionsheet));

import React, { forwardRef, memo } from 'react';
import { StyleSheet } from 'react-native';
import Backdrop from '../Backdrop';
import { Slide } from '../Transitions';
import { FocusScope } from '@react-native-aria/focus';
import { useControllableState, usePropsResolution } from '../../../hooks';
import { ModalContext } from './Context';
import Box from '../../primitives/Box';
import type { IModalProps } from './types';
import { Fade } from '../../composites/Transitions';
[...]

PR #5039, opened 2022-05-25
feat: add useRNModalOnAndroid prop to actionsheets and modals

markfuechec:
## Summary

On Android, Native Base uses the React Native Modal, which causes the rest of an application to no longer respond to user interaction. This PR adds a useRNModalOnAndroid prop to actionsheets and modals, which when set to false, causes those components to use the custom Native Base Modal.

https://github.com/GeekyAnts/NativeBase/issues/5021

## Changelog

[Android] [Added] - add useRNModalOnAndroid prop to actionsheets and modals

### DIFF
--- a/src/components/composites/Actionsheet/Actionsheet.tsx
+++ b/src/components/composites/Actionsheet/Actionsheet.tsx
@@ -13,6 +13,7 @@ const Actionsheet = (
     isOpen,
     disableOverlay,
     onClose,
+    useRNModalOnAndroid,
     ...resolvedProps
   } = usePropsResolution('Actionsheet', props);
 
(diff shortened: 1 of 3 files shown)


Viraj-10:
We have provided a similar feature in `v3.4.11`

yfeleke:
Android Actionsheet with useRNModal=true or false still doesn't fix it. The rest of the screen is always unusable. disableOverlay doesn't make  a difference either.

Irfanwani:
There is one more issue with the native base modals. Redux store is not accessible inside the components which are using inside a modal (on iOS only). After adding useRNModal, it was fixed.

Issue #5041, opened 2022-05-27
Modal avoidKeyboard not work


jonahcui:
### Description

when I open the example that uses a input inside a modal and set avoidKeyboard. The input component will be  overlay by keyboard. 

### CodeSandbox/Snack link

https://docs.nativebase.io/modal#h3-modal-with-avoidkeyboard

### Steps to reproduce

1. Go to '[Documention](https://docs.nativebase.io/modal#h3-modal-with-avoidkeyboard)' 
2. Click on 'run ios' 

### NativeBase Version

3.4

### Platform

- [ ] Android
- [ ] CRA
- [X] Expo
- [X] iOS
- [ ] Next

Viraj-10:
This issue has been fixed in NativeBase `v3.4.7`.

adamryvola:
The behaviour is still the same. I tested that on Android and iOS too.

amawaziny:
It's not working on version 3.4.25, I tested it on android expo go

carlos19960601:
same issue in NativeBase 3.4.25

holmesjr:
This issue still exists with Android and native-base 3.4.25

leciric:
I made this as a work around while it is not available.


Added a new hook to the project, that returns if the keyboard is open or not.

```tsx
import { useEffect, useState } from 'react';
import { Keyboard } from 'react-native';

export function useKeyboard() {
  const [isKeyboardVisible, setKeyboardVisible] = useState(false);

  useEffect(() => {
    const keyboardDidShowListener = Keyboard.addListener(
      'keyboardDidShow',
      () => {
        setKeyboardVisible(true);
      },
    );

    const keyboardDidHideListener = Keyboard.addListener(
      'keyboardDidHide',
      () => {
        setKeyboardVisible(false);
      },
    );

    return () => {
      keyboardDidHideListener.remove();
      keyboardDidShowListener.remove();
    };
  }, []);

  return isKeyboardVisible;
}

```

Then using this hook I added a padding bottom of 50% to the modal style

```tsx
const isKeyboardOpen = useKeyboard();
```

```tsx
<Modal
  isOpen={isModalOpen}
  onClose={() => handleCloseModal()}
  pb={isKeyboardOpen ? '50%' : '0'}
>
```

Seems to work just fine, tested using expo go.

JosephLu2022:
+1 3.25 doesn't work still

jnariai:
A workaround is to set RNModal to false
[...]

PR #5047, opened 2022-05-30
fix: usekeyboard-bottom-inset

Viraj-10:
- fix for issue #5041

### DIFF
--- a/src/utils/useKeyboardBottomInset.ts
+++ b/src/utils/useKeyboardBottomInset.ts
@@ -9,7 +9,7 @@ export const useKeyboardBottomInset = () => {
     function onKeyboardChange(e: KeyboardEvent) {
       if (
         e.startCoordinates &&
-        e.endCoordinates.screenY < e.startCoordinates.screenY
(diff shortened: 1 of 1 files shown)


holmesjr:
Yeah, this didn't work at all.

src/utils/useKeyboardBottomInset.ts
```diff
@@ -9,7 +9,7 @@ export const useKeyboardBottomInset = () => {
     function onKeyboardChange(e: KeyboardEvent) {
       if (
         e.startCoordinates &&
-        e.endCoordinates.screenY < e.startCoordinates.screenY
+        e.endCoordinates.screenY <= e.startCoordinates.screenY
```
helloworld-Andrew, line 12: ../node_modules/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactRootView.java
image
This is the createKeyboardEventPayload source code from react native. It does not contain the startCoordinates field. So that is avoidKeyboard the reason for not working on the Android platform.

Issue #4399, opened 2021-12-10
Native TextInput does not work properly when wrapped with Factory


acollazomayer:
**Describe the bug**
When using Factory creator for native text input. When the input gets focused it automatically gets unfocused

**To Reproduce**
1 - Use the text input factory example. Click on the text input and the input get unfocused automatically.

**Expected behaviour**
Text input shouldn't be unfocused.

**CodeSandBox/Snack link**
I am using the example provided in the docs. https://docs.nativebase.io/next/nativebase-factory#using-ref

**Screenshots**

https://user-images.githubusercontent.com/9849229/145554317-910b0ce2-3d58-424d-8e33-0b380ceae586.mp4

**Platform**

- NativeBase version [3.2.2]
- Device: [Simulator]
- OS: [15.0]
- Browser version [e.g. 22]
- React Native version [0.66.3]
- Expo sdk version [e.g. 41]
- node [14.15.1]
- yarn [1.22.11]

**Additional information**

acollazomayer:
I was able to make it work, mimicking part of the implementation of Factory

`import React, {memo, forwardRef} from 'react';
import {TextInput} from 'react-native';
import {
  usePropsWithComponentTheme,
  useStyledSystemPropsResolver,
} from 'native-base';
import inputTheme from '../theme/components/input';

const MaskedInput = (props, ref) => {
  const calculatedProps = usePropsWithComponentTheme(inputTheme, props);
  const [style, restProps] = useStyledSystemPropsResolver(calculatedProps);

  return <TextInput ref={ref} style={style} {...restProps} />;
};

export default memo(forwardRef(MaskedInput));
`

I think that there is something not working properly on the implementation of Factory that is making components unmount and mount again. I did not have time to investigate further. But it also happened something similar to other components

Fifaldi:
We have similar issue, on Android the Native Base Input wrapped in HOC with Factory 
`export default Factory(ContainerWithButton);`
results in the keyboard being automatically dismissed after typing a character.
Although this seems to be more happening on Android than on iOS
Simply removing the Factory wrapper from the HOC component resolves issue so agree it has to be some problem with Factory itself.
```
<ContainerWithButton
            onPress={() => navigation.navigate('onboarding')}
            buttonText={t('next', {ns: 'common'})}
            isButtonDisabled={!selectedLanguage}>
     <Input/>
</ContainerWithButton>
```
This example works only if we remove the Factory wrapper from the HOC component given above.

[...]

PR #5066, opened 2022-06-06
fix: factory fn memo for makestyle component

Viraj-10:
- fix for #4399


chornos13:
is it really doesn't has testing?

Issue #5106, opened 2022-06-25
Expose/export Tooltip interface


mathias-berg:
### Description

I would like to be able to import the type for Tooltip to make my own component fully typed.

### Problem Statement

There is not way today to get the interface or type for the Tooltip component since it doesn't seem to be exported like other types are.

### Proposed Solution or API

Export the Tooltip type like you export other types today.

Krithikj20:
Hi @mathias-berg , Thank you for reaching out to us. We will look into it.

Viraj-10:
We have released this in `v3.4.11`

PR #5172, opened 2022-07-16
fix: type-export Export the type of Tooltip component

Gilbishkosma:
Fixed the issue  #5106 (Expose/export Tooltip interface)

I have exported the type of Tooltip
`export { ITooltipProps } from './types';`


src/components/composites/Tooltip/index.ts
```diff
@@ -1 +1,2 @@
 export { Tooltip } from './Tooltip';
+export { ITooltipProps } from './types';
```
Viraj-10, line 2: this should be exported as type.

Viraj-10 (changes requested):

src/components/composites/Tooltip/index.ts
Gilbishkosma: okay, i will update it

Gilbishkosma: I have exported it as type now

PR #5180, opened 2022-07-18
fix: popover break on falsy children value

ankit-tailor:
`PopoverContent`  was breaking over the `falsy`, `null` or `string` value of children

### DIFF
--- a/src/components/composites/Popover/PopoverContent.tsx
+++ b/src/components/composites/Popover/PopoverContent.tsx
@@ -53,10 +53,7 @@ export const PopoverContent = React.forwardRef(
     let arrowElement = null;
     const restChildren: any = [];
     React.Children.toArray(props.children).forEach((child: any) => {
-      if (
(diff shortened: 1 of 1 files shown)


src/components/composites/Popover/PopoverContent.tsx
```diff
@@ -52,8 +52,11 @@ export const PopoverContent = React.forwardRef(
 
     let arrowElement = null;
     const restChildren: any = [];
-    React.Children.forEach(props.children, (child) => {
-      if (child.type.displayName === 'PopperArrow') {
+    React.Children.toArray(props.children).forEach((child: any) => {
+      if (
+        typeof child === 'object' &&
```
Viraj-10, line 57: Visit this line of code.

Viraj-10, line 57: Revisit this line code.

Viraj-10 (changes requested):

PR #5182, opened 2022-07-18
fix: falsy child break

ankit-tailor:
ButtonGroup was breaking for `false` , `null` and `string` values of children.

### DIFF
--- a/src/components/primitives/Button/ButtonGroup.tsx
+++ b/src/components/primitives/Button/ButtonGroup.tsx
@@ -27,41 +27,44 @@ export default memo(
       if (Array.isArray(children)) {
         computedChildren = React.Children.map(
           children,
-          (child: JSX.Element, index: number) => {
-            return React.cloneElement(child, {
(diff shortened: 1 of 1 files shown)


src/components/primitives/Button/ButtonGroup.tsx
```diff
@@ -27,41 +27,44 @@ export default memo(
...
-                  : { borderLeftRadius: borderRadius }
-                : {}),
-              ...(isAttached && index === children?.length - 1
-                ? direction === 'column'
-                  ? { borderBottomRadius: borderRadius }
-                  : { borderRightRadius: borderRadius }
-                : {}),
+                // when buttons are attached, remove rounded corners of all buttons except extreme buttons
+                ...(isAttached ? { borderRadius: 0 } : {}),
+                ...(isAttached && index === 0
```
Looooong, line 41: We need to take into consideration the possibility of the first child element being falsy. If that's the case then the button's corners will not be properly set.

PR #5266, opened 2022-08-11
Fix/version update rnw 0.18

Viraj-10:
- Support for `react-native-web@v0.18`, `Expo-SDK 45/46`, `react@18`,`react-native@0.69/0.70`.


dohomi:
@Viraj-10 as mentioned in https://github.com/necolas/react-native-web/issues/2363#issuecomment-1211035445 this PR is a very fragile fix and might break sooner than later. Dripsy and Restyle are handling responsive styles very well, maybe you can adopt their API for native-base in a similar way?

efoken:
@dohomi Dripsy and Restyle are indeed very good, but both do not support Server Side Rendering, while NativeBase does, because of this styles implementation. IMO this is a big advantage over other libraries.

dohomi:
@efoken I'm not comparing libraries here, I just try to point out that it is an anti pattern and not a future proof solution how this PR handles the incompatibility with react-native-weg

intergalacticspacehighway:
yo, can you refer to [this solution](https://github.com/intergalacticspacehighway/react-native-responsive-query/pull/14)? Less fragile but adds an optional [extra step for SSR](https://github.com/intergalacticspacehighway/react-native-responsive-query/tree/rnw-0.18#ssr-with-nextjs). If this step is not added, things will still work but styles won't be rendered on the server side. Would've done a PR but caught up with other issues. 😬 

Also, this approach is not backward compatible so would need a major release. It uses the CSS class approach instead of dataSet prop. https://twitter.com/nishanbende/status/1528410157801058305

Viraj-10:
Hi @dohomi, This is a temporary fix we pushed out for people who want to try rnw 0.18 and expo SDK 46. We are working on a fix which doesn't rely on the internal implementation of rnw.

PR #5304, opened 2022-08-26
Fix/colormode fix

ankit-tailor:


package.json
```diff
@@ -196,6 +196,7 @@
     "@react-stately/tabs": "3.0.0-alpha.1",
     "@react-stately/toggle": "3.2.1",
     "@types/lodash.has": "^4.5.6",
+    "@types/use-subscription": "^1.0.0",
```
Viraj-10, line 199: Is this a reliable package? take a review from @surajahmed.

Viraj-10 (changes requested):
not sure about new package `use-subscription`.

package.json
ankit-tailor: Yes, These are just types for `use-subscription`. `use-subscription` is already installed.

Issue #5307, opened 2022-08-28
Cannot customize component theme for `Fab`


Swiftwork:
### Description

Assumption is that it should be possible to override styles only on Fab and not for all Buttons.

### CodeSandbox/Snack link

https://snack.expo.dev/@swiftwork/theme-override-issue

### NativeBase Version

3.4.13

### Platform

- [X] Android
- [X] CRA
- [X] Expo
- [X] iOS
- [X] Next

Viraj-10:
Hi @Swiftwork, Thanks for reporting the issue. We will look into it.

Swiftwork:
Based on the answer from discord

> Hey, There is typo in theme for Fab. Actually you need to pass FAB. You can create an issue and track there. We'll update you on the same /Ankit Tailor

Part of the issue was the usage of `Fab` rather than `FAB` for override. However, there is still the issue that states such as `_hover` and `_pressed` don't override for the `Fab` component. I have updated the snack to include this.

Viraj-10:
Hi @Swiftwork, `_pressed` is working and you have written `_hovered` that's why hover is not working. use `_hover` it will work.

Swiftwork:
@Viraj-10 My apologies, I corrected the `_hovered` -> `_hover` typo but I must be seeing something different than you. When hovering or pressing the `FAB` button in the bottom right corner it turns blue (primary) rather than the overridden pink (secondary) rules.

Viraj-10:
Hi @Swiftwork, Thanks for that this should be happing. got it we will do the `hotfix`.

ankit-tailor:
Hi @Swiftwork, Thanks for reporting the issue. This is an architectural problem. It needs some architectural level changes. We are working on it. 

FAB uses `Button` internally. so if you want to add a variant for FAB you can use the below workaround for now.

```js
Button: {
  variants: {
    solid: {
      _light: {
        _hover: {
          bg: 'amber.500',
        },
      },
    },
    fabSolid: {
      _light: {
        _hover: {
          bg: 'red.500',
        },
        bg: 'blue.400',
      },
    },
  },
},
```

Swiftwork:
[...]

PR #5315, opened 2022-08-31
fix: fab extend theme pseudo prop

Viraj-10:
- Fab component issue. https://github.com/GeekyAnts/NativeBase/issues/5307

### DIFF
--- a/src/components/composites/Fab/Fab.tsx
+++ b/src/components/composites/Fab/Fab.tsx
@@ -1,7 +1,7 @@
 import React, { memo, forwardRef } from 'react';
 import { Button } from '../../primitives/Button';
 import type { IFabProps } from './types';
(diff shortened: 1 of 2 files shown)


Swiftwork:
@Viraj-10 can you expose/export `useComponentThemeResolver` so developers can use it to extend existing `native-base` components? I for example extend `Button` to create a `ToggleButton` and would like to treat and theme it the same way as other `native-base` components 😄

Viraj-10:
Hi @Swiftwork, Thanks for showing interest but this will be an internal hook for now since API for this hook is subject to change according to the needs.  Also, we are re-thinking this approach.

Swiftwork:
@Viraj-10 Thanks for letting me know. That's unfortunate but understandable.

Swiftwork:
@surajahmed How come this was closed? Are you taking another approach?

PR #5329, opened 2022-09-08
feat: expose the internal hooks and context for Radio variant

tvadiep:
## Summary
We're considering implement Radio variants but there're some hooks & context wasn't disposed. So I created this PR to export them out. If there's any problem, please let me know and fix it timely. Thank you very much!

## Changelog
I has export `useHasResponsiveProps` in `hooks` and `wrapStringChild`, `RadioContext` and some necessary type, hooks and context that make the re-implementation easier.

[CATEGORY] [TYPE] - Message

### DIFF
--- a/src/components/primitives/Pressable/index.tsx
+++ b/src/components/primitives/Pressable/index.tsx
@@ -1,2 +1,3 @@
 export { default as Pressable } from './Pressable';
 export type { IPressableProps } from './types';
+export * from './Pressable';
--- a/src/components/primitives/Pressable/index.tsx
+++ b/src/components/primitives/Pressable/index.tsx
@@ -1,3 +1,3 @@
 export { default as Pressable } from './Pressable';
 export type { IPressableProps } from './types';
-export * from './Pressable';
+export { useFocus, useHover, useIsPressed } from './Pressable';
--- a/src/components/primitives/index.ts
+++ b/src/components/primitives/index.ts
@@ -48,7 +48,7 @@ export type { ISpinnerProps } from './Spinner';
 export { default as Heading } from './Heading';
 export type { IHeadingProps } from './Heading';
 
-export * from './Pressable';
+export { useFocus, useHover, useIsPressed } from './Pressable';
 export type { IPressableProps } from './Pressable';
 
 export { default as Flex, Spacer } from './Flex';
--- a/src/components/primitives/index.ts
+++ b/src/components/primitives/index.ts
@@ -12,8 +12,14 @@ export type { IInputProps } from './Input';
 export { Checkbox } from './Checkbox';
 export type { ICheckboxProps, ICheckboxGroupProps } from './Checkbox';
 
-export { Radio } from './Radio';
-export type { IRadioProps, IRadioGroupProps, IRadioValue } from './Radio';
+export { Radio, RadioContext } from './Radio';
+export type {
+  IRadioProps,
+  IRadioGroupProps,
+  IRadioValue,
+  IRadioContext,
+  IRadioComponentType,
+} from './Radio';
 
 export { Icon, createIcon } from './Icon';
 export type { IIconProps } from './Icon';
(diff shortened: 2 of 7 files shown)


src/components/primitives/Pressable/index.tsx
```diff
@@ -1,2 +1,3 @@
 export { default as Pressable } from './Pressable';
 export type { IPressableProps } from './types';
+export * from './Pressable';
```
Viraj-10, line 3: Please replace `*` with only the hooks you need in `Pressable`. So in future, if we export something it shouldn't get exported with `*`.

src/components/primitives/index.ts
```diff
@@ -42,7 +48,7 @@ export type { ISpinnerProps } from './Spinner';
 export { default as Heading } from './Heading';
 export type { IHeadingProps } from './Heading';
 
-export { Pressable } from './Pressable';
+export * from './Pressable';
```
Viraj-10, line 51: same should be applied here.

src/hooks/index.ts
```diff
@@ -13,6 +13,7 @@ export {
 // TODO: Have to properly remove these as we're no longer using them internally.
 export { useThemeProps, usePropsWithComponentTheme } from './useThemeProps';
 // TODO: instead export only this.
+export * from '../hooks/useHasResponsiveProps';
```
Viraj-10, line 16: replace `*` with the intended name.

src/utils/index.ts
```diff
@@ -13,6 +13,8 @@ export {
   mergeRefs,
   composeEventHandlers,
 };
+export * from './wrapStringChild';
```
Viraj-10, line 16: replace `*` with the intended name.

src/utils/index.ts
```diff
@@ -13,6 +13,8 @@ export {
...
+export * from '../components/primitives/Radio/RadioGroup';
```
Viraj-10, line 17: replace `*` with the intended name.

Viraj-10 (changes requested):
Can you explain what's the use case for `RadioContext` and `useFormControlContext`?

src/components/primitives/Pressable/index.tsx
tvadiep: Thank you for your feedback. I fixed it in the recent commit.

src/components/primitives/index.ts
tvadiep: Thank you for your feedback. I fixed it in the recent commit.

src/hooks/index.ts
tvadiep: Thank you for your feedback. I fixed it in the recent commit.

src/utils/index.ts
tvadiep: Thank you for your feedback. I fixed it in the recent commit.

src/utils/index.ts
tvadiep: Thank you for your feedback. I fixed it in the recent commit.

tvadiep:
> Can you explain what's the use case for `RadioContext` and `useFormControlContext`?

G'day @Viraj-10, I saw that your team is using these two in your implementation. Although it's just a dummy hook/context but in the your team may have new implementations in the future and it will cause unexpected and elusive issues. To avoid that situation, we want to use these too.

Viraj-10:
> > Can you explain what's the use case for `RadioContext` and `useFormControlContext`?
> 
> G'day @Viraj-10, I saw that your team is using these two in your implementation. Although it's just a dummy hook/context but in the your team may have new implementations in the future and it will cause unexpected and elusive issues. To avoid that situation, we want to use these too.

Hi @tvadiep, We understand your requirement but be aware that this context and hooks are prone to change in future as per requirement.

Viraj-10 (approved):

src/hooks/index.ts
```diff
@@ -13,6 +13,7 @@ export {
 // TODO: Have to properly remove these as we're no longer using them internally.
 export { useThemeProps, usePropsWithComponentTheme } from './useThemeProps';
 // TODO: instead export only this.
+export { useHasResponsiveProps } from '../hooks/useHasResponsiveProps';
```
rayan1810, line 16: This is a deprecated hook.

rayan1810: We are working on a new function that provides the same functionality, will release it shortly.

Issue #5429, opened 2022-09-28
Factory - Typescript - Expression produces a union type that is too complex to represent. ts(2590)


Kivylius:
### Description

There seems to be an issues with typescript using a factory for a native base components in VSCode. The error is union type that is too complex. Snack link seesm to be ok, but as you can see, it not checking types correctly (see prop `unknownprop` not displaying as error) 

### CodeSandbox/Snack link

https://snack.expo.dev/ANiYSN7p8?platform=web

### Steps to reproduce

1. Go to snack link 
2. Click on download 
3. Scroll down to load into VSCode or VsCodium
4. See error typescript about union types (but code works fine)

<img width="1171" alt="Screen Shot 2022-09-28 at 12 35 11" src="https://user-images.githubusercontent.com/1041335/192768907-92945829-299b-404f-8688-6e3234873645.png">

The snack for expo seems to not report typescript errors correctly, as you can see the unknown prop is not displaying correctly.

I have tried several ways with same issues:

<img width="835" alt="Screen Shot 2022-09-28 at 12 36 17" src="https://user-images.githubusercontent.com/1041335/192769108-eabd1d9d-0872-4e42-8658-ab927b3ac48b.png">

<img width="1162" alt="Screen Shot 2022-09-28 at 12 36 46" src="https://user-images.githubusercontent.com/1041335/192769394-7b5699f3-3806-4a08-9a6e-e3a522089d25.png">

It seems to work ok with native components:

<img width="958" alt="Screen Shot 2022-09-28 at 12 36 54" src="https://user-images.githubusercontent.com/1041335/192769244-316d8719-2620-412b-ace5-f0467f72931e.png">

### NativeBase Version

^3.4.17

### Platform

- [ ] Android
- [ ] CRA
- [ ] Expo
- [ ] iOS
- [ ] Next

### Other Platform

VsCode

### Additional Information

N/a

Viraj-10:
Hi @Kivylius, Thank you for reporting this and creating a PR. we will look into it.

PR #5431, opened 2022-09-28
fix: issues with union types for Factory

Kivylius:
fixes union type of typescript #5429

## Summary

The issues is identified in #5429
When using ReactNative component in a factory, the typescript is not working as expected. The code works correctly but the typescript is the issues.

## Changelog

[CATEGORY] [Fix] - fixes union type of typescript for Factory.

## Test Plan

Ive ran the test and there seems to be no braking changes. Both Jest and Typescript run fines. I believe it's a bug.

### DIFF
--- a/src/factory/types.ts
+++ b/src/factory/types.ts
@@ -1,7 +1,6 @@
 import type { PlatformProps } from '../components/types';
 import type { StyledProps } from '../theme/types';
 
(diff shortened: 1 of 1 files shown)


Viraj-10:
Hi @Kivylius, Thank you for reporting this and creating a PR. we will look into it.

PR #5504, opened 2022-10-21
Saeed

SaeedPoureshghi:
## Summary
on `Actionsheet` when user disable overlay by adding `disableOverlay` it is impossible to close with click outside of modal. 
With this change you can disable overlay and prevent from dimming screen and close action menu easily.


SaeedPoureshghi:
not applicable

PR #5512, opened 2022-10-27
fix(Tabs): uncomment export tabs

lobor:
## Summary
No export of Tabs component

## Changelog

[General] [Fixed] - uncomment export tabs


Viraj-10:
Hi @lobor, Tab is Alpha and it's not stable right now. We are working on releasing this component soon. Thanks for the PR.

PR #5534, opened 2022-11-04
fix(checkbox):passing accessibilityHint prop to the component

mfarhankasmani:
## Summary
NativeBase allow user to pass accessbilityHint for checkbox (ICheckboxProps) but accessibilityHint prop is not being extracted. Currently Hint is not being announce and this PR resolves that bug.

## Changelog

[CATEGORY] [TYPE] - Message


rayan1810 (approved):
Looks good to me.

PR #5559, opened 2022-11-16
feat: added optional support for angle

### LINKED ISSUE #5538: angle support on react-native-linear-gradient integration
### Description

`react-native-linear-gradient` [supports passing an angle instead of start and end positions](https://github.com/react-native-linear-gradient/react-native-linear-gradient#useangle--angle--anglecenter) to calculate gradients. The current integration of `LinearGradient` inside NativeBase does not support that.

### Problem Statement

The current implementation strictly uses [`start` and `end`](https://github.com/GeekyAnts/NativeBase/blob/5c52b856b5ce4c71eaaed8db03bedd853c6b0ba0/src/components/primitives/Box/index.tsx#L83), but `react-native-linear-gradient` already supports angle usage.

### Proposed Solution or API

Add support for `useAngle`, `angle` and `angleCenter` props in `linearGradient` style configuration. Since `lgrad` isn't typed anyway, we could just extract those props from it and pass it to `LinearGradient`:

```typescript
return (
      <Gradient
        ref={ref}
        {...safeAreaProps}
        colors={lgrad.colors}
        start={lgrad.useAngle ? null : startObj}
        end={lgrad.useAngle ? null : endObj}
        useAngle={lgrad.useAngle}
        angle={lgrad.angle}
        angleCenter={lgrad.angleCenter}
        locations={lgrad.locations}>
        {/* {React.Children.map(children, (child) =>
          typeof child === 'string' || typeof child === 'number' ? (
            <Text {..._text}>{child}</Text>
          ) : (
            child
          )
        )} */}
        {wrapStringChild(children, _text)}
      </Gradient>
    );
```

This can be much cleaner, but that's the gist of this request.

### Alternatives

_No response_

### Additional Information

_No response_

proohit:
## Summary

Adds (optional) support for `useAngle` `angle` and `angleCenter` props for `LinearGradient`.

closes #5538 

## Changelog

[General] [Added] - Add support for angle props in `LinearGradient`

## Test Plan

I've added a storybook example

```jsx
export const ExampleWithAngle = () => {
  return (
    <Box
      bg={{
        linearGradient: {
          colors: ['lightBlue.300', 'violet.800'],
          angle: 45,
          angleCenter: { x: 0.5, y: 0.5},
        },
      }}
      p="12"
      w="72"
      rounded="xl"
      _text={{
        fontSize: 'md',
        fontWeight: 'medium',
        color: 'warmGray.50',
        textAlign: 'center',
      }}
    >
      This is a Box with Linear Gradient using angle props
    </Box>
  );
}
```

### DIFF
--- a/example/storybook/stories/components/primitives/Box/LinearGrad.tsx
+++ b/example/storybook/stories/components/primitives/Box/LinearGrad.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
 import { Box } from 'native-base';
+import React from 'react';
 export const Example = () => {
   return (
     <Box
@@ -24,3 +24,28 @@ export const Example = () => {
     </Box>
   );
 };
+
+export const ExampleWithAngle = () => {
+  return (
+    <Box
+      bg={{
+        linearGradient: {
+          colors: ['lightBlue.300', 'violet.800'],
+          angle: 45,
+          angleCenter: { x: 0.5, y: 0.5},
+        },
+      }}
+      p="12"
+      w="72"
+      rounded="xl"
+      _text={{
+        fontSize: 'md',
+        fontWeight: 'medium',
+        color: 'warmGray.50',
+        textAlign: 'center',
+      }}
+    >
+      This is a Box with Linear Gradient using angle props
+    </Box>
+  );
+}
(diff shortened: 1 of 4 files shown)


douglasscriptore:
I need this :/

proohit:
Let's hope for @surajahmed or team to pick it up:)

proohit:
I just realized that this does **not** work with `expo-linear-gradient`, since they use an own implementation without `react-native-linear-gradient`. The storybook examples however use that, so may not be able to view in storybook. Any suggestions are appreciated

PR #5648, opened 2023-01-19
fix: use "base" fontFamily in web

GuilleDF:
## Summary

The fontFamily prop generated currently looks like `FontName-MediumItalic`. This is generally how it's expected by react-native, but not how a browser would expect it. On web, the font family is generally only `FontName` and other props such as `fontWeight` are used in conjunction. This disables the logic to generate this "compound" fontFamily for web.

## Changelog

[Web] [Changed] - Use normal fontFamily in web

## Test Plan

I ran the testbed and checked the Text stories to see if something broke.

### DIFF
--- a/src/hooks/useResolvedFontFamily.web.ts
+++ b/src/hooks/useResolvedFontFamily.web.ts
@@ -1,0 +1,18 @@
+import type { ITheme } from '../theme';
+/**
+ *
+ * @param props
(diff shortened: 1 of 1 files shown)


GuilleDF:
Update: I added an example for testing, but feel free to discard it if it's not useful :)

PR #5685, opened 2023-03-01
Removes selection prop on select input on safari

### LINKED ISSUE #5111: Select element takes focus from input element on react native web on Safari browser
### Description

Only a single character is accepted on controlled input when a Select element is present

### CodeSandbox/Snack link

https://snack.expo.dev/MOBQ7fNEN

### Steps to reproduce

1. On Safari browser
2. In Expo snack, type a letter in input component - shows up
3. Type another letter - It doesn't show up

### NativeBase Version

3.4.7

### Platform

- [ ] Android
- [ ] CRA
- [X] Expo
- [ ] iOS
- [ ] Next

### Other Platform

Web - Safari

### Additional Information

I think it is the internal input element of a Select element which takes focus.

gfviegas:
Fixes #5111

## Summary
When a Select is present on a form rendered on Safari (both iOS and Mac tested), every change in ANY input (even others than the Select) causes the screen to focus on the select input and loses focus on the changing input.

This PR aims to fix #5111 . There's a minimal reproduction of the problem this PR solves in this issue.

### DIFF
--- a/src/components/primitives/Select/Select.tsx
+++ b/src/components/primitives/Select/Select.tsx
@@ -154,6 +154,8 @@ const Select = (
     'shadow',
     'opacity',
   ]);
+  
+  const isSafari = Platform.OS === 'web' && window.safari
 
   const commonInput = (
     <Input
@@ -172,6 +174,7 @@ const Select = (
       isDisabled={isDisabled}
       pointerEvents="none"
       variant={variant}
+      selection={isSafari ? undefined : nonLayoutProps.selection}
     />
   );
 


src/components/primitives/Select/Select.tsx
```diff
@@ -154,6 +154,8 @@ const Select = (
     'shadow',
     'opacity',
   ]);
+  
+  const isSafari = Platform.OS === 'web' && window.safari
```
gfviegas, line 158:
```suggestion
  const isSafari = Platform.OS === 'web' && navigator.userAgent.match(/Version\/[\d.]+.*Safari/)
```

Actually, in iPhone, the `safari` property is not set.

mikevercoelen:
Good catch! This should also be applied to `Input` component.

robbporto:
Any chance that this can be merged, please? I'm having to use an old version of native-base because of this bug and this PR fixes it.

Issue #5778, opened 2023-08-02
SSRProvider warning


muhamed-misbeer:
### Description

In React 18, SSRProvider is not necessary and is a noop. You can remove it from your app. 

### CodeSandbox/Snack link

-

### Steps to reproduce

2. Click on '...' 
3. Scroll down to '...' 

### NativeBase Version

3.4

### Platform

- [X] Android
- [ ] CRA
- [X] Expo
- [X] iOS
- [ ] Next

deepbag:
nativebase need to remove <SSRProvider>{children}</SSRProvider> from src/core/NativeBaseProvider.tsx

darrylyoung:
This is a duplicate of #5758. There's a lot more information in the other issue.

PR #5764, opened 2023-07-05
SSRProvider Issue Fix

kurucaner:
## Changelog

[CATEGORY] [TYPE] - Message

PR #5788, opened 2023-10-05
Fix SSRProvider issue on React 18+

Lev-Shapiro:
## Summary

I've just started using React Native and chose NativeBase as UI library, however it seems that it has bug for the version 18+ of React, which couple people mentioned in Issues and even made PRs, however they were closed for some unknown reason and the error is still there, so I've made this pull request in order to solve the SSRProvider bug ( terminal: ` WARN  In React 18, SSRProvider is not necessary and is a noop. You can remove it from your app.  ` )

### DIFF
--- a/src/core/NativeBaseProvider.tsx
+++ b/src/core/NativeBaseProvider.tsx
@@ -94,7 +94,7 @@ const NativeBaseProvider = (props: NativeBaseProviderProps) => {
             <OverlayProvider isSSR>
               <ToastProvider>
                 <InitializeToastRef />
-                <SSRProvider>{children}</SSRProvider>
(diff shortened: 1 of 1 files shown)


Lev-Shapiro:
Inspired to do this PR by: #5788, #5764 and #5778

jetonk:
@Lev-Shapiro After I applied the changed into `src/core/NativeBaseProvider.tsx` same as in your PR, it looks it doesn't want to work.



Screenshot 2024-01-24 at 12 22 56

leksansilva (approved):

WillianSugiyama (approved):

Jhonatanmizu (approved):

